PR #28, opened 2018-02-23
Update DefaultLicensingSecurityCacheRuleInvalidator#invalidate

aubincleme:
* The invalidator now handle extensions installed at farm level
* Fix bug locking the wiki when a paying app is installed at farm level

### DIFF
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
@@ -50,6 +50,8 @@ import com.xpn.xwiki.XWikiContext;
 @Singleton
 public class DefaultLicensingSecurityCacheRuleInvalidator implements LicensingSecurityCacheRuleInvalidator
 {
+    private static final String WIKI_NAMESPACE = "wiki:";
+
     /**
      * Fair read-write lock to suspend the delivery of cache updates while there are loads in progress.
      */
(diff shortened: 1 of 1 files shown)


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
```diff
@@ -94,25 +100,46 @@ public void invalidate(XarInstalledExtension extension)
...
+     * @param extension the extension to invalidate
+     * @param wikiReference the wiki to invalidate
+     */
+    private void invalidateForWiki(XarInstalledExtension extension, WikiReference wikiReference)
+    {
+        for (XarEntry entry : extension.getXarPackage().getEntries()) {
+            securityCache.remove(
+                    securityReferenceFactory.newEntityReference(
+                            documentReferenceResolver.resolve(entry, wikiReference)
+                    )
```
mflorea: Why double indentation?

mflorea:
Hi @aubincleme, the proposed changes make sense, but could you please report an issue and describe the steps to reproduce and how the problem manifests itself. For instance I just installed the OnlyOffice Connector on farm and I didn't notice any problem.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
```diff
@@ -94,25 +100,46 @@ public void invalidate(XarInstalledExtension extension)
...
                 }
+            } else {
+                // As the extension is installed at farm level, we have to invalidate every cache entry for this
+                // extension in every wiki.
+                for (String wikiId : wikiDescriptorManager.getAllIds()) {
+                    invalidateForWiki(extension, new WikiReference(wikiId));
+                }
             }
+        } catch (WikiManagerException e) {
+            // Fail silently
```
Enygma2002: You should never fail silently when an exception is thrown. It clearly signals a problem and it should be handled.

Normally, the exception should be thrown up the execution chain, however, the current method does not throw an exception. It would be recommended that it does or you could even wrap it into an unchecked IllegalStateException and still throw it. Worst case, you should at least log the exception (as an error).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
```diff
@@ -94,25 +100,46 @@ public void invalidate(XarInstalledExtension extension)
...
-                        securityCache.remove(
-                            securityReferenceFactory.newEntityReference(
-                                documentReferenceResolver.resolve(entry, wikiRef)
-                            )
-                        );
+            Collection<String> namespaces = extension.getNamespaces();
+            if (namespaces != null) {
+                for (String namespace : namespaces) {
+                    if (namespace.startsWith("wiki:")) {
+                        invalidateForWiki(extension, new WikiReference(namespace.substring(5)));
```
Enygma2002: Even though this is not necessarily part of the issue addressed, it would be nice to replace that magic string and magic number with a string constant (e.g. `WIKI_NAMESPACE`) and a method call to get its length (i.e. `WIKI_NAMESPACE.length()`).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/DefaultLicensingSecurityCacheRuleInvalidator.java
aubincleme: Logging the error would be better IMO

aubincleme:
@mflorea I described the steps to reproduce the issue in #29 FYI

PR #93, opened 2021-02-11
Fixed quality build

oanalavinia:
* added unit tests to meet the expected 0.38 coverage ratio

### DIFF
--- a/application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
@@ -20,16 +20,23 @@
 package com.xwiki.licensing.internal;
 
 import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.never;
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import javax.inject.Provider;
+
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.xwiki.component.util.DefaultParameterizedType;
 import org.xwiki.configuration.ConfigurationSource;
+import org.xwiki.environment.Environment;
 import org.xwiki.test.mockito.MockitoComponentMockingRule;
 
 import com.xwiki.licensing.LicensingConfiguration;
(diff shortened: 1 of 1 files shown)


application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
```diff
@@ -75,4 +96,22 @@ public void getBlocklistWithEmptyList() throws Exception
 
         assertEquals(Collections.emptyList(), mocker.getComponentUnderTest().getAutoUpgradeBlocklist());
     }
+
+    @Test
+    public void getLocalStorePath() throws Exception
+    {
+        when(this.configurationSource.getProperty("licensing.localStorePath")).thenReturn("path");
+        this.mocker.getComponentUnderTest().getLocalStorePath();
```
mflorea, line 104: Why don't you assert the return value?

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
```diff
@@ -75,4 +96,22 @@ public void getBlocklistWithEmptyList() throws Exception
...
+
+        verify(this.environment, never()).getPermanentDirectory();
```
mflorea, line 106: ``verify`` is normally used for **void** methods which is not the case here. I'm not sure it's worth verifying if ``getPermanentDirectory`` is called. I would simply assert the value returned by ``getLocalStorePath``.

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
```diff
@@ -75,4 +96,22 @@ public void getBlocklistWithEmptyList() throws Exception
...
+    }
+
+    @Test
+    public void getLocalStorePathWithNullProperty() throws Exception
+    {
+        when(this.configurationSource.getProperty("licensing.localStorePath")).thenReturn(null);
+        this.mocker.getComponentUnderTest().getLocalStorePath();
```
mflorea, line 113: Same as above.

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensingConfigurationTest.java
```diff
@@ -75,4 +96,22 @@ public void getBlocklistWithEmptyList() throws Exception
...
+
+        verify(this.environment).getPermanentDirectory();
```
mflorea, line 115: Why not use ``when`` to return some value and thus make sure that the permanent directory is actually used (the return value might as well be discarded).

mflorea (changes requested):

mflorea (approved):

PR #94, opened 2021-02-23
Generate application license automatically from the Extension Manager

oanalavinia:
* added a listener for extension install; if the installed extension is a visible paid extension (not dependency of another paid extension) and the owner information is complete filled up, then request a trial license and update the licenses


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensingConfiguration.java
```diff
@@ -102,4 +110,34 @@ public File getLocalStorePath()
         }
     }
 
+    @Override
+    public String getStoreTrialURL()
+    {
+        return this.storeConfig.getProperty("storeTrialURL");
```
oanalavinia, line 116: @mflorea Should we provide here a default value, like "http://url", to avoid an exception while building the trial url here https://github.com/xwikisas/application-licensing/pull/94/files#diff-cfa828b50efb9640c7d353996f2d315d4074ef6b5835174fc0cdfc8560e6df08R113 ?
Same for getStoreUpdateURL()

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -112,4 +114,25 @@
...
+    {
+        Collection<ExtensionId> paidExtensions = getLicensedExtensions();
+
+        Set<ExtensionId> visiblePaidExtensions = new HashSet<ExtensionId>(paidExtensions);
+        for (ExtensionId extensionId : paidExtensions) {
+            InstalledExtension installedExtension = installedExtensionRepository.getInstalledExtension(extensionId);
+
+            Collection<ExtensionDependency> dependencies = installedExtension.getDependencies();
+            for (ExtensionId paidExtension : paidExtensions) {
+                if (dependencies.stream().filter(o -> o.getId().contentEquals(paidExtension.getId())).findFirst()
```
mflorea, line 129:
```suggestion
                if (dependencies.stream().filter(dependency -> dependency.getId().equals(paidExtension.getId())).findFirst()
```

Regarding ``contentEquals``, I don't think it's needed when both sides are Strings.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensingConfiguration.java
mflorea: I don't think there is a good default value here (and same for ``getStoreUpdateURL()``) so I think the code that calls these methods should check if the return value is empty/null and don't do anything if it's the case (e.g. don't get a license automatically).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensingConfiguration.java
```diff
@@ -102,4 +110,34 @@ public File getLocalStorePath()
...
+    @Override
+    public String getStoreUpdateURL()
+    {
+        return this.storeConfig.getProperty("storeUpdateURL");
+    }
+
+    @Override
+    public String getLicensingOwnerFirstName()
+    {
+        return this.ownerConfig.getProperty("firstName", "");
```
mflorea, line 128: I'm not sure the empty string is useful as default value. Isn't that already the case if you don't specify a default value?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -0,0 +1,95 @@
...
+
+                if (getTrialResponse.contains("error")) {
+                    logger.info("Failed to add trial license");
+                } else {
+                    logger.info("Added trial license");
+                    getTrialLicenseHandler.updateLicenses();
+                }
+            }
+        } catch (Exception e) {
+            logger.info("Failed to get trial license for [{}]. Root cause is [{}]", extension.getId(),
```
mflorea, line 91:
```suggestion
            logger.warn("Failed to get trial license for [{}]. Root cause is [{}]", extension.getId(),
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -0,0 +1,95 @@
...
+                ExceptionUtils.getRootCause(e));
```
mflorea, line 92:
```suggestion
                ExceptionUtils.getRootCauseMessage(e));
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -0,0 +1,95 @@
...
+                String getTrialResponse =
+                    getTrialLicenseHandler.getURLContent(getTrialLicenseHandler.getTrialURL(extension.getId()));
+
+                if (getTrialResponse.contains("error")) {
+                    logger.info("Failed to add trial license");
+                } else {
+                    logger.info("Added trial license");
+                    getTrialLicenseHandler.updateLicenses();
+                }
+            }
```
mflorea, line 89: All these ``isOwnerDataComplete``, ``isVisibleLicensedExtension``, ``getURLContent``, ``getTrialURL`` and ``updateLicenses`` are internal methods that don't need to be exposed as API. I would rather have a single ``generateTrialLicense(extensionId)`` method in a ``TrialLicenseGenerator`` component, that does all the work.

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensingConfiguration.java
oanalavinia: The _null_ value is returned if no default value is given, leading to a NPE here https://github.com/xwikisas/application-licensing/pull/94/files/6261250232b15a50ff58bdb04459690455f09f15#diff-cfa828b50efb9640c7d353996f2d315d4074ef6b5835174fc0cdfc8560e6df08R199

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
oanalavinia: So `TrialLicenseGenerator` should replace `GetTrialLicenseHandler` and be placed inside the internal package?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensingConfiguration.java
mflorea: You could use ``StringUtils.isEmpty()`` there to avoid the NPE.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
mflorea: Yes. You probably need only two methods: ``canGenerateTrialLicense(extensionId)`` and ``generateTrialLicense(extensionId)``.

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/GetTrialLicenseListenerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public MockitoComponentMockingRule<GetTrialLicenseListener> mocker =
+        new MockitoComponentMockingRule<>(GetTrialLicenseListener.class);
+
+    TrialLicenseGenerator trialLicenseGenerator;
+
+    DefaultInstalledExtension extension;
+
+    ExtensionId extensionId;
+
+    URL trialUrl;
```
mflorea, line 54: This is not needed.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+     * Generate trial license for the given extension.
+     *
+     * @param extensionId the extension for which to generate a trial license
+     */
+    public void generateTrialLicense(ExtensionId extensionId)
+    {
+        try {
+            URL trialURL = getTrialURL(extensionId);
+            if (trialURL == null) {
+                logger.info(FAILED_TO_ADD_TRIAL_LICENSE);
```
mflorea, line 108: I would use debug level here and give move details: "Failed to add trial license because the licensor configuration is not complete. Check your store trial URL and owner details."

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+                return;
+            }
+
+            XWikiContext xcontext = contextProvider.get();
+            String getTrialResponse = xcontext.getWiki().getURLContent(trialURL.toString(), xcontext);
+
+            if (getTrialResponse.contains("error")) {
+                logger.info(FAILED_TO_ADD_TRIAL_LICENSE);
```
mflorea, line 116: I would use debug level here also with "Failed to generate trial license on store."

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+            } else {
+                logger.info("Added trial license");
```
mflorea, line 118: Also debug level, and maybe reverse "Trial license added for [{}]." with extension id.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * Check if the given extension is a mandatory licensed extension and the licensing owner information is complete.
+     *
+     * @param extensionId extension to be checked
+     * @return true if a trial license can be generated for the given extension, false otherwise
+     */
+    public Boolean canGenerateTrialLicense(ExtensionId extensionId)
+    {
+        return isOwnerDataComplete() && isMandatoryLicensedExtension(extensionId);
```
mflorea, line 135: Shouldn't we also check if there is already an active license for the specified extension? Otherwise we will try to generate a trial license when upgrading a paid app for which there is a license already. No?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Retrieve license updates from the XWiki Store.
+     */
+    @SuppressWarnings("unchecked")
+    private void updateLicenses()
+    {
+        try {
+            URL licensesUpdateURL = getLicensesUpdateURL();
+            if (licensesUpdateURL == null) {
+                return;
```
mflorea, line 176: Maybe log a debug level message.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -0,0 +1,251 @@
...
+     * @return the URL for updating licenses, or null if it cannot be constructed
+     * @throws URISyntaxException if the URL is not valid
+     * @throws MalformedURLException if an error occurred while constructing the URL
+     */
+    private URL getLicensesUpdateURL() throws URISyntaxException, MalformedURLException
+    {
+        String storeUpdateURL = licensingConfig.getStoreUpdateURL();
+        // In case the property has no filled value, the URL cannot be constructed.
+        if (storeUpdateURL == null) {
+            logger.warn("Failed to update licenses");
```
mflorea, line 208: I would move this up in ``updateLicenses()`` as debug level.

mflorea (changes requested):
Almost there :)

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -105,17 +103,18 @@ public void generateTrialLicense(ExtensionId extensionId)
...
 
             XWikiContext xcontext = contextProvider.get();
             String getTrialResponse = xcontext.getWiki().getURLContent(trialURL.toString(), xcontext);
 
             if (getTrialResponse.contains("error")) {
-                logger.info(FAILED_TO_ADD_TRIAL_LICENSE);
+                logger.debug("Failed to generate trial license on store.");
             } else {
-                logger.info("Added trial license");
+                logger.debug(String.format("Trial license added for %s", extensionId.getId()));
```
mflorea, line 117:
```suggestion
                logger.debug("Trial license added for [{}].", extensionId.getId());
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -105,17 +103,18 @@ public void generateTrialLicense(ExtensionId extensionId)
         try {
             URL trialURL = getTrialURL(extensionId);
             if (trialURL == null) {
-                logger.info(FAILED_TO_ADD_TRIAL_LICENSE);
+                logger.debug("Failed to add trial license because the licensor configuration is not complete. "
+                    + "Check your store trial URL and owner details.");
```
mflorea, line 107: Would be good to specify the extension id.

```suggestion
                logger.debug("Failed to add trial license for [{}] because the licensor configuration is not complete. "
                    + "Check your store trial URL and owner details.", extensionId.getId());
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -105,17 +103,18 @@ public void generateTrialLicense(ExtensionId extensionId)
...
                 return;
             }
 
             XWikiContext xcontext = contextProvider.get();
             String getTrialResponse = xcontext.getWiki().getURLContent(trialURL.toString(), xcontext);
 
             if (getTrialResponse.contains("error")) {
-                logger.info(FAILED_TO_ADD_TRIAL_LICENSE);
+                logger.debug("Failed to generate trial license on store.");
```
mflorea, line 115: Same here.

mflorea (approved):

Issue #98, opened 2021-05-04
The auto-generated trial license can onverride an existing license


oanalavinia:
This issue is described on the initial opened bug from [xwiki-pro repository](https://github.com/xwikisas/xwiki-pro/issues/22)

Issue #97, opened 2021-05-04
Trial license is generated for the paid apps dependencies which should be covered by the parent license.


oanalavinia:
When installing a paid app, the trial license is automatically generated. If this app has as dependency another paid app (like Meeting application has XPoll application as a dependency), it's license should be covered by the parent license. The problem is that right now a new trial license is generated automatically for the paid app dependency as well.

The problem is discussed in [this](https://github.com/xwikisas/application-licensing/issues/92#issuecomment-829815012) comment as well.

mflorea:
Maybe we should listen to "extension install job finished" rather than on ``ExtensionInstalled``, i.e. to run our code after the entire install plan was executed (all dependencies installed).

PR #99, opened 2021-05-05
The auto-generated trial license can override an existing license #98

oanalavinia:
Trial license is generated for the paid apps dependencies which should be covered by the parent license. #97
* try to generate a trial license only after the extension is fully installed; this will assure that all the present licenses are validated and there is no trial license generated for the paid apps dependencies, since they would not be seen as mandatory paid apps
* update licenses for the instance before checking if a trial license is needed

### DIFF
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
@@ -27,10 +27,15 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.extension.Extension;
 import org.xwiki.extension.ExtensionDependency;
 import org.xwiki.extension.ExtensionId;
 import org.xwiki.extension.InstalledExtension;
+import org.xwiki.extension.ResolveException;
+import org.xwiki.extension.repository.ExtensionRepositoryManager;
 import org.xwiki.extension.repository.InstalledExtensionRepository;
 import org.xwiki.job.event.JobFinishedEvent;
 import org.xwiki.observation.EventListener;
@@ -57,6 +62,12 @@ public class GetTrialLicenseListener implements EventListener
 
     @Inject
     private InstalledExtensionRepository installedExtensionRepository;
+
+    @Inject
+    private ExtensionRepositoryManager repositoryManager;
+
+    @Inject
+    private Logger logger;
 
     @Override
     public List<Event> getEvents()
@@ -95,13 +106,17 @@ public class GetTrialLicenseListener implements EventListener
             trialLicenseGenerator.generateTrialLicense(extensionId);
         } else {
             InstalledExtension installedExtension = installedExtensionRepository.getInstalledExtension(extensionId);
+
             if (installedExtension != null) {
                 Collection<ExtensionDependency> dependencies = installedExtension.getDependencies();
-
                 for (ExtensionDependency dependency : dependencies) {
-                    ExtensionId dependencyId =
-                        new ExtensionId(dependency.getId(), dependency.getVersionConstraint().getVersion());
-                    tryGenerateTrialLicenseRecursive(dependencyId);
+                    try {
+                        Extension extension = repositoryManager.resolve(dependency);
+                        tryGenerateTrialLicenseRecursive(extension.getId());
+                    } catch (ResolveException e) {
+                        logger.warn("Failed to check [{}] for a license. Root cause is [{}]", dependency.getId(),
+                            ExceptionUtils.getRootCauseMessage(e));
+                    }
                 }
             }
         }
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
@@ -30,12 +30,10 @@ import javax.inject.Singleton;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
-import org.xwiki.extension.Extension;
 import org.xwiki.extension.ExtensionDependency;
 import org.xwiki.extension.ExtensionId;
 import org.xwiki.extension.InstalledExtension;
 import org.xwiki.extension.ResolveException;
-import org.xwiki.extension.repository.ExtensionRepositoryManager;
 import org.xwiki.extension.repository.InstalledExtensionRepository;
 import org.xwiki.job.event.JobFinishedEvent;
 import org.xwiki.observation.EventListener;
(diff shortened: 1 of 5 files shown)


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -250,4 +249,18 @@ private Boolean isOwnerDataComplete()
...
+    /**
+     * Check if the given paid extension has already a license. We need to retrieve license updates from the XWiki Store
+     * before, in case there is already an active license of the extension.
+     *
+     * @param extensionId the extension to be checked
+     * @return true if the extension is licensed, false otherwise.
+     */
+    private Boolean isLicensed(ExtensionId extensionId)
+    {
+        updateLicenses();
```
mflorea, line 262: Will this slow down the extension install/upgrade? I hope it doesn't break the install/upgrade when the XWiki Store is down. What I don't like is that:
* you update all licenses: maybe this is needed because the installed extension could be covered by a license generated for another extension, in which case a comment would be good
* you can make this call multiple times during an install if the installed extension has licensed dependencies (which will be installed first); I'm stating to think that listening to ``ExtensionInstalled`` is not the best option, it would be better to listen to "extension install job finished" and then update all licenses (once) and then look for extensions that are missing a license and try to generate one.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -65,10 +66,14 @@ public String getName()
...
     {
-        DefaultInstalledExtension extension = (DefaultInstalledExtension) source;
+        List<ExtensionId> extensions = ((JobFinishedEvent) event).getRequest().getProperty("extensions");
+        // Retrieve license updates to be sure that we don't override an existing license.
+        trialLicenseGenerator.updateLicenses();
 
-        if (trialLicenseGenerator.canGenerateTrialLicense(extension.getId())) {
-            trialLicenseGenerator.generateTrialLicense(extension.getId());
+        for (ExtensionId extensionId : extensions) {
+            if (trialLicenseGenerator.canGenerateTrialLicense(extensionId)) {
```
mflorea, line 74: Although it's not common, we can have a free extension that depends on a paid extension. AFAIK, the list of extensions you get from the job request represents only the top level extensions that the user has requested to be installed explicitly. It doesn't include the transitive dependencies. So you need to go through the dependency tree (starting from these top level extensions) to look for licensed extensions. Of course, you don't go "down" when you find a licensed extension (its dependencies will be covered).

mflorea (changes requested):
BTW @oanalavinia , do we fail safe when a trial license cannot be generated because the user already extended the trial license once? I mean this use case:

* you install a paid app, a trial license is generated
* you extend the trial license
* trial license expires
* you upgrade the paid app and a new trial license is attempt but should fail safely

oanalavinia:
> BTW @oanalavinia , do we fail safe when a trial license cannot be generated because the user already extended the trial license once? I mean this use case:
> 
> * you install a paid app, a trial license is generated
> * you extend the trial license
> * trial license expires
> * you upgrade the paid app and a new trial license is attempt but should fail safely

Yes, it tries to generate a trial licence only when the given extension is unlicensed `License.UNLICENSED.equals(licensorProvider.get().getLicense(extensionId))`

mflorea:
Right, but what if the user removes the (trial and expired) licenses from the file system, restarts the server and then upgrades a paid app? The deleted (and expired) licenses will be re-downloaded by ``updateLicenses`` so we're safe, right?

oanalavinia:
Right

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -65,10 +73,27 @@ public String getName()
...
 
-        if (trialLicenseGenerator.canGenerateTrialLicense(extension.getId())) {
-            trialLicenseGenerator.generateTrialLicense(extension.getId());
+            for (ExtensionDependency dependency : dependencies) {
+                ExtensionId dependencyId =
+                    new ExtensionId(dependency.getId(), dependency.getVersionConstraint().getVersion());
+                if (trialLicenseGenerator.canGenerateTrialLicense(dependencyId)) {
+                    trialLicenseGenerator.generateTrialLicense(dependencyId);
+                }
+            }
```
mflorea, line 96: You're checking only one level in the dependency tree but we could have licensed extensions many levels below. I think you need to iterate **all** the dependency **tree**, making sure to not iterate the descendants of a node for which you generate a license. Also, you shouldn't treat the root node differently. I would write a generic method that receives an extension id as input (the root) and then call itself recursively for its child nodes if we don't generate a license for the input extension id.

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -78,20 +78,30 @@ public void onEvent(Event event, Object source, Object data)
...
+                Collection<ExtensionDependency> dependencies = installedExtension.getDependencies();
 
-            for (ExtensionDependency dependency : dependencies) {
-                ExtensionId dependencyId =
-                    new ExtensionId(dependency.getId(), dependency.getVersionConstraint().getVersion());
-                if (trialLicenseGenerator.canGenerateTrialLicense(dependencyId)) {
-                    trialLicenseGenerator.generateTrialLicense(dependencyId);
+                for (ExtensionDependency dependency : dependencies) {
+                    ExtensionId dependencyId =
+                        new ExtensionId(dependency.getId(), dependency.getVersionConstraint().getVersion());
```
mflorea, line 103: I'm not sure about this part. The version constraint specified in the POM for the dependency doesn't have to match the version that is actually installed (used at runtime). For instance, extension A might declare in its POM a dependency on extension B with version 5.3 (which means it needs **at least** version 5.3) and at runtime work with B v7.1 (because this version is needed by some other extension for instance). Moreover, you could have an extension C that specifies in its POM that B is a feature (it covers B, e.g. if B was renamed to C) and thus you can have A work with C, a different extension id than the one specified in the POM as dependency.

So I think you need to ``resolve`` the dependency id into an installed extension (i.e. to see which installed extension satisfies that dependency).

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -95,13 +106,17 @@ public void tryGenerateTrialLicenseRecursive(ExtensionId extensionId)
...
+
             if (installedExtension != null) {
                 Collection<ExtensionDependency> dependencies = installedExtension.getDependencies();
-
                 for (ExtensionDependency dependency : dependencies) {
-                    ExtensionId dependencyId =
-                        new ExtensionId(dependency.getId(), dependency.getVersionConstraint().getVersion());
-                    tryGenerateTrialLicenseRecursive(dependencyId);
+                    try {
+                        Extension extension = repositoryManager.resolve(dependency);
```
mflorea, line 114: I think you should use ``installedExtensionRepository`` to resolve the dependency for two reasons:

* we want to know which **installed** extension satisfies that dependency
* we're not interested in **core** licensed extensions so we should only look in the installed extension repository

mflorea (changes requested):

mflorea (approved):

PR #103, opened 2021-07-23
Update parent version to 11.10 #101

oanalavinia:
* modify parent and ignore revapi errors
* export pages from xwiki 11.10
* move the template of extensionAutoUgraded event to delete the need of a wiki page, since there are two events with the same template
* update tests and readme
* removed code needed for avoiding dead lock on xwiki versions older than 10.4RC1


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/notifications/ExtensionAutoUpgradedEventDisplayer.java
```diff
@@ -56,44 +53,45 @@
...
+            // Set a document in the context to act as the current document when the template is rendered.
+            xcontext.setDoc(xcontext.getWiki().getDocument(LICENSOR_DOC, xcontext));
+
+            // Bind the event to some variable in the velocity context.
+            scriptContext.setAttribute(EVENT_BINDING_NAME, eventNotification, ScriptContext.ENGINE_SCOPE);
 
-            Template customTemplate = templateManager.createStringTemplate(
-                displayerObj.getStringValue("notificationTemplate"), displayerDoc.getAuthorReference());
             return (customTemplate != null) ? this.templateManager.execute(customTemplate)
                 : this.templateManager.execute("notification/default.vm");
```
mflorea, line 92: The fallback on the default template means that a missing custom template is somehow valid (no a problem). But we bundle the custom template in the same JAR as this code so it should always be found, unless there's a serious problem, in which case we should be aware of. I think the fallback make sense when this code doesn't "own" the custom template, i.e. when the custom template is provided by some other module and it _may_ be missing. And since the custom template is own by this code I wouldn't do the fallback. Unless this is a best practice for notification displayers that I'm not aware of.

mflorea: To be clear, the fallback doesn't hurt and can be considered as bulletproofing. My issue is that it sends the wrong message that the custom template _may_ be missing, while it should _never_ be missing since it is bundled with the code.

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/extensionAutoUpgraded/extensionAutoUpgraded.vm
```diff
@@ -0,0 +1,53 @@
...
+
+#macro (displayEventDetails $event)
+  #set ($userName = $event.user.name)
+  <tr>
+    <td>
+      <span class="notification-event-user">$userName</span>
+    </td>
+    <td class="description">$event.getBody()</td>
+    <td class="text-right text-muted">$escapetool.xml($datetool.whenIs($event.date))</td>
+</tr>
```
mflorea, line 17: Bad alignment.

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/extensionAutoUpgraded/extensionAutoUpgraded.vm
```diff
@@ -0,0 +1,53 @@
...
+
+#set ($mainIcon = "bell")
+#set ($smallIcon = "arrow_up")
+#set ($compositeEventDate = $escapetool.xml($datetool.whenIs($event.dates.get(0))))
+
+#macro (displayEventDetails $event)
+  #set ($userName = $event.user.name)
+  <tr>
+    <td>
+      <span class="notification-event-user">$userName</span>
```
mflorea, line 13: If it doesn't need to be XML escaped then I would add a comment explaining why (does it contain a link to the user profile?).

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/extensionAutoUpgraded/extensionAutoUpgraded.vm
```diff
@@ -0,0 +1,53 @@
+#template('notification/macros.vm')
+
+#includeMacros('Licenses.Code.VelocityMacros')
+
+#set ($mainIcon = "bell")
+#set ($smallIcon = "arrow_up")
+#set ($compositeEventDate = $escapetool.xml($datetool.whenIs($event.dates.get(0))))
+
+#macro (displayEventDetails $event)
+  #set ($userName = $event.user.name)
```
mflorea, line 10: Seems to be used only once below, so I'm not sure it's worth having a variable (due to the way Velocity works I'm trying to avoid creating unneeded variables which may influence some other code that uses the same name in the same namespace / context).

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/extensionAutoUpgraded/extensionAutoUpgraded.vm
```diff
@@ -0,0 +1,53 @@
...
+  <tr>
+    <td>
+      <span class="notification-event-user">$userName</span>
+    </td>
+    <td class="description">$event.getBody()</td>
```
mflorea, line 15: I assume the event body can have HTML?

mflorea (commented):
Looks good with some minor comments. Thanks!

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/extensionAutoUpgraded/extensionAutoUpgraded.vm
oanalavinia: The event body is just a  simple text message added  https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/notifications/ExtensionAutoUpgradedEventConverter.java#L61
which is constructed in here https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/UpgradeExtensionHandler.java#L118

mflorea: So it seems it doesn't contain HTML, at least not ATM, thus it would be safer I think, for now, to escape the special XML characters. Users can mess with translations, putting values we don't control, but this shouldn't break the XWiki UI.

mflorea (approved):

Issue #115, opened 2022-01-10
Wiki creation job freezes at 75% on an instance with multiple subwikis and pro applications installed


oanalavinia:
**Prerequisites:** 
1. Have Licensor API 1.21 installed on farm
2. Have multiple sub wikis (e.g. 5), each with at least a pro application installed (e.g. Diagram, Meeting Application (Pro) )

**Steps to reproduce:**
1. Create a new subwiki `test6`
2. Observe the wiki creation job progress 

**Expected result:**
After waiting for a reasonable amount of time, the wiki creation job progress displays the message `The wiki test6 has been successfully created. You can now access it.`  

**Actual result:**
Even after a long time, the wiki creation job is still running 
![wikiCreation](https://user-images.githubusercontent.com/22794181/148768208-d917907b-e431-4584-9140-7adc257c3a9f.png)
and the progress remains at 0.7499 even after 2h of waiting
```
<jobStatus xmlns="http://www.xwiki.org">
<link href="http://localhost:8081/xwiki/rest/jobstatus/wikicreation/createandinstall/test6" rel="self"/>
<id>wikicreation/createandinstall/test6</id>
<state>RUNNING</state>
<progress>
<offset>0.7499999999999989</offset>
<currentLevelOffset>0.7499999999999989</currentLevelOffset>
</progress>
<startDate>2022-01-10T14:23:08.572+02:00</startDate>
<serialized>true</serialized>
<isolated>true</isolated>
</jobStatus>
```

Note: this issue is related to #92 and can be reproduced starting with Licensing version 1.17

gdelhumeau:
For the record, it was also breaking the upgrade of existing subwikis.
Thanks for having closed this!

PR #117, opened 2022-01-13
Wiki creation job freezes at 75% on an instance with multiple subwikis and pro applications installed #115

oanalavinia:
* refactor GetTrialLicenseListener#tryGenerateTrialLicenseRecursive to remember the dependencied there were already verified
* consider the namespace of the verified extensions
* modified TrialLicenseGenerator#canGenerateTrialLicense to have the least costly conditions first
* refactor DefaultLicensedExtensionManager#getMandatoryLicensedExtensions to check also transitive dependencies efficiently by remembering which dependencies were already verified
* cache the list of mandatory licensed extensions and invalidate it after each extension job
* update and add unit tests

### DIFF
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
@@ -123,18 +123,21 @@ public class DefaultLicensedExtensionManager implements LicensedExtensionManager
     @Override
     public Set<ExtensionId> getMandatoryLicensedExtensions()
     {
-        Set<ExtensionId> mandatoryLicensedExtensions = computeMandatoryLicensedExtensions();
-        this.cachedMandatoryLicensedExtensions = mandatoryLicensedExtensions;
-        return mandatoryLicensedExtensions;
+        Set<ExtensionId> mandatoryLicensedExtensions = this.cachedMandatoryLicensedExtensions;
+        if (mandatoryLicensedExtensions == null) {
+            mandatoryLicensedExtensions = computeMandatoryLicensedExtensions();
+            this.cachedMandatoryLicensedExtensions = mandatoryLicensedExtensions;
+        }
+        return Collections.unmodifiableSet(mandatoryLicensedExtensions);
     }
 
     private synchronized Set<ExtensionId> computeMandatoryLicensedExtensions()
     {
-        if (this.cachedMandatoryLicensedExtensions != null) {
-            return this.cachedMandatoryLicensedExtensions;
+        Set<ExtensionId> mandatoryLicensedExtensions = this.cachedMandatoryLicensedExtensions;
+        if (mandatoryLicensedExtensions != null) {
+            return mandatoryLicensedExtensions;
         }
 
-        Set<ExtensionId> mandatoryLicensedExtensions = new HashSet<ExtensionId>();
         Collection<ExtensionId> allLicensedExtensions = getLicensedExtensions();
         // Extensions for which it was verified if the dependencies contain licensed extensions.
         Set<ExtensionId> verifiedExtensions = new HashSet<ExtensionId>();
@@ -172,7 +175,7 @@ public class DefaultLicensedExtensionManager implements LicensedExtensionManager
             ExtensionId dependencyId = installedDependency.getId();
             mandatoryLicensedExtensions.remove(dependencyId);
 
-            if (verifiedExtensions.contains(installedDependency.getId())) {
+            if (verifiedExtensions.contains(dependencyId)) {
                 continue;
             }
             verifiedExtensions.add(dependencyId);
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
@@ -123,9 +123,9 @@ public class DefaultLicensedExtensionManager implements LicensedExtensionManager
     @Override
     public Set<ExtensionId> getMandatoryLicensedExtensions()
     {
-        Set<ExtensionId> cache = computeMandatoryLicensedExtensions();
-        this.cachedMandatoryLicensedExtensions = cache;
-        return cache;
+        Set<ExtensionId> mandatoryLicensedExtensions = computeMandatoryLicensedExtensions();
+        this.cachedMandatoryLicensedExtensions = mandatoryLicensedExtensions;
+        return mandatoryLicensedExtensions;
     }
 
     private synchronized Set<ExtensionId> computeMandatoryLicensedExtensions()
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
@@ -123,14 +123,22 @@ public class DefaultLicensedExtensionManager implements LicensedExtensionManager
     @Override
     public Set<ExtensionId> getMandatoryLicensedExtensions()
     {
+        Set<ExtensionId> cache = computeMandatoryLicensedExtensions();
+        this.cachedMandatoryLicensedExtensions = cache;
+        return cache;
+    }
+
+    private synchronized Set<ExtensionId> computeMandatoryLicensedExtensions()
+    {
         if (this.cachedMandatoryLicensedExtensions != null) {
             return this.cachedMandatoryLicensedExtensions;
         }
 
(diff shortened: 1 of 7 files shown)


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/TrialLicenseGenerator.java
```diff
@@ -238,7 +238,7 @@ private URL getLicensesUpdateURL() throws URISyntaxException, MalformedURLExcept
     private Boolean isMandatoryLicensedExtension(ExtensionId extensionId)
     {
         return licensedExtensionManager.getMandatoryLicensedExtensions().stream()
-            .filter(extension -> extension.getId().contentEquals(extensionId.getId())).findFirst().isPresent();
+            .anyMatch(mandatoryExtensionId -> mandatoryExtensionId.equals(extensionId));
```
mflorea, line 241: Couldn't you simply use ``contains``?

    return licensedExtensionManager.getMandatoryLicensedExtensions().contains(extensionId);

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManagerTest.java
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Unit tests for {@link DefaultLicensedExtensionManager}.
+ * 
+ * @version $Id$
+ */
+public class DefaultLicensedExtensionManagerTest
+{
+    @Rule
+    public MockitoComponentMockingRule<DefaultLicensedExtensionManager> mocker =
+        new MockitoComponentMockingRule<>(DefaultLicensedExtensionManager.class);
```
mflorea, line 55: Couldn't you use Junit5? See https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/ .

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -60,6 +60,11 @@
      */
     public static final String LICENSOR_EXTENSION_ID = "com.xwiki.licensing:application-licensing-licensor-api";
 
+    /**
+     * Cache the list of installed licensed extensions that are not covered by the license of another extension.
+     */
+    private Set<ExtensionId> cachedMandatoryLicensedExtensions;
```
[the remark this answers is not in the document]
oanalavinia: @mflorea Would be the right fix to use `@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)`? This would make possible to add a state.
Or it is better to maybe use a cache component for saving the extensions list?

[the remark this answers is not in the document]
mflorea: If ``DefaultLicensedExtensionManager`` stops being a singleton then the cache doesn't matter that much anymore. Each component that injects ``LicensedExtensionManager`` will have its own instance with its own cache, which basically means that the list of mandatory licensed extensions is going to be recomputed each time.

I think it needs to remain a singleton but you need to make sure that whenever its state (``cachedMandatoryLicensedExtensions``) is read (``getMandatoryLicensedExtensions``) the return value is valid. The state can become invalid through writes, so you need to make sure that after each write the state is valid. If multiple writes are needed to go from one valid state to another valid state then you need to make sure those writes cannot be interleaved with a read from a different thread. This is usually done using synchronized blocks/methods or semaphores.

In you case the problems are:
* ``cachedMandatoryLicensedExtensions`` can be in a partial state when read (because some other thread is computing the cache) -> you can fix this by setting ``cachedMandatoryLicensedExtensions`` only at the end when you have the final list of mandatory licensed extensions
* multiple threads can compute the cache at the same time; only one thread should be allowed to do this; you can move the code that computes the cache in a **synchronized** method
* some thread can invalidate the cache (set to null) while another one is computing the case; you can fix this with the first item above + returning the computed value instead of ``this.cachedMandatoryLicensedExtensions`` in ``getMandatoryLicensedExtensions``. Basically:
  ```
  Set<ExtensionId> cache = computeMandatoryLicensedExtensions();
  this.cachedMandatoryLicensedExtensions = cache;
  return cache;
  ```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -116,23 +121,69 @@
     }
 
     @Override
-    public Collection<ExtensionId> getMandatoryLicensedExtensions()
+    public Set<ExtensionId> getMandatoryLicensedExtensions()
+    {
+        Set<ExtensionId> mandatoryLicensedExtensions = computeMandatoryLicensedExtensions();
+        this.cachedMandatoryLicensedExtensions = mandatoryLicensedExtensions;
+        return mandatoryLicensedExtensions;
```
mflorea, line 128: I think we can avoid synchronization if the mandatory licensed extensions are already cached.

```suggestion
        Set<ExtensionId> mandatoryLicensedExtensions = this.cachedMandatoryLicensedExtensions;
        if (mandatoryLicensedExtensions == null) {
            mandatoryLicensedExtensions = computeMandatoryLicensedExtensions();
            this.cachedMandatoryLicensedExtensions = mandatoryLicensedExtensions;
        }
        return Collections.unmodifiableSet(mandatoryLicensedExtensions);
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -116,23 +121,69 @@
...
+    }
+
+    private synchronized Set<ExtensionId> computeMandatoryLicensedExtensions()
+    {
+        if (this.cachedMandatoryLicensedExtensions != null) {
+            return this.cachedMandatoryLicensedExtensions;
+        }
```
mflorea, line 135: The risk is that some other thread can set the ``cachedMandatoryLicensedExtensions`` to ``null`` between the test and the return.

```suggestion
        Set<ExtensionId> mandatoryLicensedExtensions = this.cachedMandatoryLicensedExtensions;
        if (mandatoryLicensedExtensions != null) {
            return mandatoryLicensedExtensions;
        }
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -116,23 +121,69 @@
...
+
+        Set<ExtensionId> mandatoryLicensedExtensions = new HashSet<ExtensionId>();
```
mflorea, line 137: You initialize it below, so you don't need to do it here.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java
```diff
@@ -116,23 +121,69 @@
...
+            InstalledExtension installedDependency =
+                this.installedExtensionRepository.getInstalledExtension(dependency.getId(), namespace);
+            if (installedDependency == null) {
+                continue;
             }
+
+            ExtensionId dependencyId = installedDependency.getId();
+            mandatoryLicensedExtensions.remove(dependencyId);
+
+            if (verifiedExtensions.contains(installedDependency.getId())) {
```
mflorea, line 175:
```suggestion
            if (verifiedExtensions.contains(dependencyId)) {
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/GetTrialLicenseListener.java
```diff
@@ -78,46 +84,58 @@ public void onEvent(Event event, Object source, Object data)
...
 
         for (ExtensionId extensionId : extensions) {
-            tryGenerateTrialLicenseRecursive(extensionId);
+            InstalledExtension installedExtension = installedExtensionRepository.getInstalledExtension(extensionId);
+            Stack<ExtensionId> dependencyPath = new Stack<>();
+            dependencyPath.push(extensionId);
+            tryGenerateTrialLicenseRecursive(dependencyPath, installedExtension.getNamespaces());
         }
+
+        licensedExtensionManager.invalidateMandatoryLicensedExtensionsCache();
```
mflorea, line 93: I'm wondering if the list of mandatory licensed extensions should be invalidated before or after the trial license generation (which needs this list). ATM it seems the trial license generation is using the old list (before the install job). Is this OK?

mflorea (changes requested):

mflorea (approved):

Issue #110, opened 2021-11-19
Notify the administrators when one of the installed licensed applications has a new version available


mflorea:
This should be done only if automatic upgrades are turned off for that application or globally. There should also be a configuration option to disable these notifications (e.g. put a checkbox on the Licenses administration section).

The notification message could indicate that in order to disable these notifications you need to:
* either use the corresponding configuration
* or enable automatic upgrades, see #109

oanalavinia:
-  Added a scheduler job that does a check for new versions available and sends a notification (one per application/wiki/version); this job is triggered once a week since usually upgrades are not done more frequently
-  The notification is sent to those that have these notifications (_New extension version available_) enabled and the System filter disabled, since the notification is send by superadmin user;

The notification:
![licensing_110](https://user-images.githubusercontent.com/22794181/187136151-143d4d7e-1730-4a6b-902d-315d072e928f.png)

The notifications settings:
![licensing_110_notifSettings](https://user-images.githubusercontent.com/22794181/187137128-c9b6461b-eeef-46c3-a8ad-d0de426ae93a.png)

PR #126, opened 2022-05-17
 Notify the administrators when one of the installed licensed applications has a new version available #110

oanalavinia:
* add `NewExtensionVersionAvailable` event and a job scheduler to send this event every week
* update `ExtensionEventConverter` to set the custom message on the event body and `LicensingSchedulerListener` to schedule / reschedule this event as well on licensing upgrade or install
* have a `NewExtensionVersionAvailableEventDisplayer` to provide a custom template for the event
* add unit tests

newVersionNotification


mflorea:
Receiving the notification that a new version of an extension is available is a plus, but the user (administrator) needs to be able to take an **action** (i.e. upgrade). We could provide links to individual extensions that the administrator could follow to perform the upgrade and / or a link to the extension updater where they can see all the available upgrades.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/notifications/newVersion/NewExtensionVersionAvailableEventDisplayer.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Inject
+    private Logger logger;
+
+    @Override
+    public Block renderNotification(CompositeEvent eventNotification) throws NotificationException
+    {
+        XWikiContext xcontext = contextProvider.get();
+        ScriptContext scriptContext = this.scriptContextManager.getScriptContext();
+        Template customTemplate = this.templateManager.getTemplate("newVersionAvailable/newVersionAvailable.vm");
```
mflorea, line 82: Is the folder ``newVersionAvailable`` really needed? Does it have multiple templates?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/notifications/newVersion/NewExtensionVersionAvailableEventDisplayer.java
```diff
@@ -0,0 +1,102 @@
...
+
+        try {
+            // Set a document in the context to act as the current document when the template is rendered.
+            xcontext.setDoc(xcontext.getWiki().getDocument(LICENSOR_DOC, xcontext));
+            // Bind the event to some variable in the velocity context.
+            scriptContext.setAttribute(EVENT_BINDING_NAME, eventNotification, ScriptContext.ENGINE_SCOPE);
+
+            return this.templateManager.execute(customTemplate);
+        } catch (Exception e) {
+            logger.warn("Failed to render custom template. Root cause is: [{}]", ExceptionUtils.getRootCauseMessage(e));
```
mflorea, line 92: 'custom template' doesn't mean much. You know which template failed to be executed.

mflorea (changes requested):

oanalavinia:
I added a link to an upgrade page corresponding to each notification
image

Also, I added an a new property to the `AutomaticUpgradesClass`, namely `newVersionNotifiedExtensions`, to keep an evidence of the already send notifications, to avoid multiplying them.
Note that I wasn't sure about using the existing `AutomaticUpgradesClass`, the name of the class might not be generic enough, but they are related.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/LicensingConfiguration.java
```diff
@@ -72,4 +72,17 @@
...
+     */
+    List<String> getNewVersionNotifiedExtensions();
+
+    /**
+     * Set the list of extensions with a new version available, that users have been notified of.
+     *
+     * @param newVersionNotifiedExtensions the list of extensions with a new version available, that users have been
+     *            notified of
+     */
+    void setNewVersionNotifiedExtensions(List<String> newVersionNotifiedExtensions);
```
mflorea, line 87: I don't see how this is related to configuration, which normally includes settings that the administrators can set to tweak the behavior of the application. In this case, I don't see why the administrators would set themselves the lists of extensions that have a new version available. This looks more like some **internal** state that the licensing application needs to manage in order to prevent duplicate notifications. The administrators shouldn't be aware of this.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -77,26 +79,42 @@ public void checkLicensedExtensionsAvailableVersions()
...
     {
         InstalledExtension installedExtension =
             installedRepository.getInstalledExtension(extensionId.getId(), namespace);
-        List<Version> versions = upgradeExtensionHandler.getInstallableVersions(installedExtension.getId());
-        if (versions.size() > 0) {
-            String message = String.format("%s - %s - %s", installedExtension.getName(),
-                namespace != null ? namespace : "root", versions.get(0));
-            this.observationManager.notify(new NewExtensionVersionAvailableEvent(), extensionId.getId(), message);
+        List<Version> installableVersions = upgradeExtensionHandler.getInstallableVersions(installedExtension.getId());
+        if (installableVersions.size() <= 0) {
```
mflorea, line 97:
```suggestion
        if (installableVersions.isEmpty()) {
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -77,26 +79,42 @@ public void checkLicensedExtensionsAvailableVersions()
...
+            return;
         }
-    }
 
+        List<String> newVersionNotifiedExtensions = new ArrayList<>(licensingConfig.getNewVersionNotifiedExtensions());
+        String namespaceName = namespace != null ? namespace : "root";
+        // Create an identified for the extension, by consider also the version and namespace.
```
mflorea, line 103:
```suggestion
        // Create an identifier for the extension, by consider also the version and namespace.
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -77,26 +79,42 @@ public void checkLicensedExtensionsAvailableVersions()
...
+        String verifiedExtensionId =
+            String.format("%s-%s-%s", extensionId.getId(), namespaceName, installableVersions.get(0));
```
mflorea, line 105: Why "verified"? And it would be good to mention that the first item in ``installableVersions`` is the most recent (right?).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -77,26 +79,42 @@ public void checkLicensedExtensionsAvailableVersions()
...
-                    notifyExtensionAvailableVersions(installedExtension.getId(), namespace);
+                    notifyExtensionAvailableVersion(installedExtension.getId(), namespace);
                 }
             }
         }
 
     }
 
-    private void notifyExtensionAvailableVersions(ExtensionId extensionId, String namespace)
+    private void notifyExtensionAvailableVersion(ExtensionId extensionId, String namespace)
```
mflorea, line 92:
```suggestion
    private void notifyExtensionVersionAvailable(ExtensionId extensionId, String namespace)
```

application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/newVersionAvailable.vm
```diff
@@ -4,13 +4,29 @@
 #set ($smallIcon = "arrow_up")
 #set ($compositeEventDate = $escapetool.xml($datetool.whenIs($event.dates.get(0))))
 
+#macro (getExtensionUpgradeURL $event)
+  ## The event body contains information about the extension id, version and targeted namespace.
+  #set ($extensionInfo = $event.getBody().split(' - '))
```
mflorea, line 9: First of all, I think the message contains the extension name not id. Thus, what happens if the extension name contains ``' - '``? This is too fragile and thus error-prone. If you want to pass structured information in the message body then why not use a standard format such as JSON or XML? Or maybe you can extract this information from elsewhere.

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/LicensingConfiguration.java
oanalavinia: What would be the best way of saving this internal state? Maybe have a separate page that holds an object with the list of extensions?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -79,42 +79,40 @@ public void checkLicensedExtensionsAvailableVersions()
...
 
-            String message =
-                String.format("%s - %s - %s", installedExtension.getName(), namespaceName, installableVersions.get(0));
-            this.observationManager
-                .notify(
-                    new NewExtensionVersionAvailableEvent(
-                        new ExtensionId(extensionId.getId(), installableVersions.get(0)), namespace),
-                    extensionId.getId(), message);
+            String message = String.format("{\"extensionName\":\"%s\", \"namespace\": \"%s\", \"version\": \"%s\"}",
+                installedExtension.getName(), namespaceName, installableVersions.get(0));
```
mflorea, line 112: We don't control the extension name and its version so they may contain special characters that break the JSON. Either escape the values or, better, use a Java map that you then serialize to JSON using the Jackson library that we bundled in XWiki.

mflorea (changes requested):

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/LicensingConfiguration.java
mflorea: First of all, keeping all the versions for all the extensions doesn't scale (if we don't do any cleanup). The list grows indefinitely. I think we could keep only the last **known** installable version for each extension and namespace (that we have notified the administrators of). Each time this last know version changes we send a new notification. So I'm thinking we could use a class with 3 fields (namespace, extension id and version) and add objects of this type to the licensing config page, making sure we have only one object per namespace and extension id. This should be "hidden" behind a component that gives you the last known installable version and allows you to update it, in case we want to implement it differently in the future.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewVersionNotificationManager.java
```diff
@@ -0,0 +1,157 @@
...
+                }
+                notificationObjectExists = true;
+            }
+        }
+        return notificationObjectExists;
+    }
+
+    private boolean objectHasValue(BaseObject obj, String name, String value)
+    {
+        return obj.getStringValue(name) != null && obj.getStringValue(name).equals(value);
```
mflorea, line 155:
```suggestion
        return Objects.equals(obj.getStringValue(name), value);
```

application-licensing-licensor/application-licensing-licensor-api/src/test/java/com/xwiki/licensing/internal/upgrades/NewVersionNotificationManagerTest.java
```diff
@@ -0,0 +1,187 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link NewVersionNotificationManager}.
+ *
+ * @version $Id$
+ * @since 1.23
+ */
+@ComponentTest
+public class NewVersionNotificationManagerTest
```
mflorea, line 55:
```suggestion
class NewVersionNotificationManagerTest
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/NewExtensionVersionAvailableManager.java
```diff
@@ -98,21 +109,25 @@ private void notifyExtensionVersionAvailable(ExtensionId extensionId, String nam
...
-            licensingConfig.setNewVersionNotifiedExtensions(newVersionNotifiedExtensions);
-
-            String message = String.format("{\"extensionName\":\"%s\", \"namespace\": \"%s\", \"version\": \"%s\"}",
-                installedExtension.getName(), namespaceName, installableVersions.get(0));
-            this.observationManager.notify(
-                new NewExtensionVersionAvailableEvent(new ExtensionId(extensionId.getId(), installableVersions.get(0)),
-                    namespace), extensionId.getId(), message);
+        try {
+            String namespaceName = namespace != null ? namespace : "root";
+            if (!this.newVersionNotificationManager.isNotificationAlreadySent(installedExtension.getName(),
```
mflorea, line 114: Shouldn't you use the extension id instead?

mflorea (approved):

Issue #131, opened 2022-11-28
New extension version available notification throws NPE on newer versions of XWiki


oanalavinia:
**Prerequisites**: have Licensor Application version 1.23 installed and follow https://github.com/xwikisas/application-licensing/issues/110#issuecomment-1229838690 to have `New extension version available` notifications enabled

**Steps to reproduce**:
1. Install a licensed extension at a lower version than the last available one (e.g. Application Diagram version 1.15.3)
2. Go to the Job Scheduler (`Scheduler.WebHome`) and observe the `New version available` job
3. Trigger the job or wait for it to be triggered at the `next fire time`
4. Go on a new page and open the new notification

**Expected result**: the notification is displayed correctly and it suggests you to upgrade to a higher version of Diagram

**Actual result**: an error is displayed inside the notification
![image](https://user-images.githubusercontent.com/22794181/204308002-e4f50089-c376-4600-b72a-971c89658a06.png)

I reproduced it on 14.4.6, but not on 11.10

PR #132, opened 2022-11-28
New extension version available notification throws NPE on newer versions of XWiki #131

oanalavinia:
* remove unnecessary code, since the wiki name is already contained in each notification message

The error was caused by the fact that `$event.document` is no longer defined by default in newer versions of xwiki. In our case, this wasn't actually even needed since there's no associated document to the event and `$event.document.wikiReference.name` was always the same on older versions of xwiki

### DIFF
--- a/application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/newVersionAvailable.vm
+++ b/application-licensing-licensor/application-licensing-licensor-api/src/main/resources/templates/newVersionAvailable.vm
@@ -34,14 +34,6 @@
 #end
 
 #define ($content)
-  #set ($document = $xwiki.getDocument($event.document))
-  <div class="notification-page">
-    ## Notifications are rendered in the context of their wiki, so we need to use the XWikiContext#originalWikiId
-    ## to actually know where the request comes from.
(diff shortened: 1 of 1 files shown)


mflorea:
> ``$event.document`` is no longer defined by default in newer versions of XWiki

Is this a regression on XWiki Standard side, or maybe it wasn't documented (i.e. unstable or internal API)?

oanalavinia:
I couldn't find the place where the document was set before.
But, from my understanding it shouldn't have been set anyway since on my implementation for converting the event  I call the [default implementation](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/notifications/ExtensionEventConverter.java#L60) which should [not set a document ](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventConverter.java#L79) since it will not reach that line (data is not a document)

Issue #136, opened 2023-01-03
Allow to extend a licence in the Licensing Manager


ldubost:
The objective is to add an "Extend" button which will prefill the information for a new license extending the current license.

PR #137, opened 2023-01-03
#136 Allow to extend a licence in the Licensing Manager

ldubost:
This code is adding a "Copy and Extend License" button allow to open a prepopulated new license page.
This allows to speed up when extending licences.

### DIFF
--- a/application-licensing-manager/application-licensing-manager-ui/src/main/resources/License/Code/Extend.xml
+++ b/application-licensing-manager/application-licensing-manager-ui/src/main/resources/License/Code/Extend.xml
@@ -1,0 +1,105 @@
+<?xml version="1.1" encoding="UTF-8"?>
+
+<!--
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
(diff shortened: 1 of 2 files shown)


application-licensing-manager/application-licensing-manager-ui/src/main/resources/License/Code/Extend.xml
```diff
@@ -0,0 +1,105 @@
...
+* Issue Date: $pageDoc.issueDate
+* Current Expiration Date: $pageDoc.expirationDate
+* Proposed new Expiration Date: $tentativeDate
+
+
+
+{{html}}
+&lt;form&gt;
+&lt;input type='hidden' name='action' value='add'/&gt;
+&lt;input type='hidden' name='extend' value='1'/&gt;
```
mflorea, line 86: I'd change the action to ``extend`` and check for "confirmation" above before doing the redirect.

```suggestion
&lt;input type='hidden' name='action' value='extend'/&gt;
&lt;input type='hidden' name='confirm' value='true'/&gt;
```

mflorea (changes requested):

mflorea:
Fixed by https://github.com/xwikisas/application-licensing/commit/822eb56f23e17d23ee628df34375c2fe645e8ff5 .

PR #159, opened 2023-12-04
legacy jquery "size()" replacted with "length"

knee-cola:
jQuery has dropped the support for `size` method in v3.0 - it is replaced by `length` property.

Current version of XWiki uses jQuery v3.6.0 , which brakes the current implementation of this plugin.

See:
* https://api.jquery.com/size/


oanalavinia:
Thanks for the contribution!

Issue #157, opened 2023-11-14
Create a missing license rendering macro


oanalavinia:
For displaying the `There is no valid license for..` message on each licensed application, right now we have 2 macros [getMissingLicenseMessage](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/VelocityMacros.xml#L114) and [getMissingLicenseMessageHTML](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/VelocityMacros.xml#L119C9-L119C37) from the `Licenses.Code.VelocityMacros` page. E.g. of usages on [diagram](https://github.com/xwikisas/application-diagram/blob/3a2858e097d76217d1c2f247b2efa9c0d31c4bd8/application-diagram-ui/src/main/resources/Diagram/WebHome.xml#L67) and [pdfViewer](https://github.com/xwikisas/macro-pdfviewer/blob/47f020adbf427d7dd882d8d057112ba7d38b31e7/macro-pdfviewer-api/src/main/resources/templates/pdfviewer/pdfviewer.vm#L140)
![image](https://github.com/xwikisas/application-licensing/assets/22794181/ac94f127-a61a-4778-a946-04249cd42f85)
The problem with this is that we need to explicitly give view right for the `VelocityMacros` page, to avoid issues like https://github.com/xwikisas/application-licensing/issues/155

The proposal is to have a rendering macro that will be registered without this view right (like it happens with translations) and use it
`{{missingLicenseMessage extensionName="diagram.extension.name" /}}`

We should also investigate a solution for the usage in pdfViewer, maybe to specify that the template uses wiki syntax by adding `##!source.syntax=xwiki/2.1`, such as in [here](https://raw.githubusercontent.com/xwiki/xwiki-platform/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/distribution/report.wiki).

oanalavinia:
This will also be of help for those that use custom groups instead of the standard `XWikiAdminGroup` and `XWikiAllGroup`. These users do not benefit from the fixes of #155 #156 (they need to specify manually the name of their custom groups)

lucaa:
[...]

PR #161, opened 2023-12-18
Create a missing license rendering macro #157

oanalavinia:
* create the `missingLicenseMessage` macro

With this, every licensed app will remove the [old velocity macro](https://github.com/xwikisas/application-diagram/blob/3a2858e097d76217d1c2f247b2efa9c0d31c4bd8/application-diagram-ui/src/main/resources/Diagram/WebHome.xml#L67) usage with this macro call, removing the need to include the `Licenses.Code.VelocityMacros` page.
Note that the [PDFViewer macro code](https://github.com/xwikisas/macro-pdfviewer/blob/47f020adbf427d7dd882d8d057112ba7d38b31e7/macro-pdfviewer-api/src/main/resources/templates/pdfviewer/pdfviewer.vm#L140) can also be updated to use this wiki macro.


application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/MissingLicenseMessageMacro.xml
```diff
@@ -0,0 +1,374 @@
...
+#end
+#set ($extensionName = $services.localization.render("${wikimacro.parameters.extensionName}"))
+#set ($translation = $services.localization.render("licensor.missingLicense", 'xwiki/2.1',
+  [$extensionName, 'LINKSTART', 'LINKEND']))
+#set ($linkLabel = $stringtool.substringBetween($translation, 'LINKSTART', 'LINKEND'))
+#set ($escapedLinkLabel = $services.rendering.escape($linkLabel, 'xwiki/2.1'))
+#set ($escapedLinkTarget = $services.rendering.escape("path:${licenseHomeURL}", 'xwiki/2.1'))
+{{error}}
+  $translation.replace("LINKSTART${linkLabel}LINKEND", "[[$escapedLinkLabel&gt;&gt;$escapedLinkTarget]]")
+{{/error}}
```
mflorea, line 262: This looks more complicated than it should be. Can we deprecate this translation key and add another one that has a single parameter, the link?

oanalavinia: I updated it. But I couldn't find a best practice for deprecating a translation. So I just deleted the old one and when applications will upgrade the licensing dependency, they have to take care of using the new one.  
What I found was only for platform https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationPropertyDeprecations , but it does not seem to be of help here

Issue #163, opened 2024-02-15
Upgrade licensor parent version to 14.10


oanalavinia:

oanalavinia:
Reopen in order to fix also the functional tests

PR #167, opened 2024-04-04
Upgrade licensor parent version to 14.10 #163

oanalavinia:
* update parent version
* export from 14.10 and remove deprecated property defaultCategory
* update jaxb plugin configuration, which was duplicated from the parent version
* add a dependency needed by jaxb
* skip banned dependency type enforcer
* update functional tests as well to use 14.10
* [XWIKI-19676](https://github.com/xwiki/xwiki-platform/commit/26728f3f23658288683667a5182a916c7ecefc52#diff-f58643b3b5f5908a02d8a292acdc2ec715f68a3c88b4faaeb7adfbf86be72722R89) added changes that will have effect on https://github.com/xwikisas/application-licensing/blob/master/application-licensing-manager/application-licensing-manager-ui/src/main/resources/License/Code/GenerateCertificates.xml#L104 . These changes mainly broke the unit tests and some changes were needed [here](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-manager/application-licensing-manager-api/src/main/java/com/xwiki/licensing/script/ManagerScriptService.java#L61) in order to use SHA256

**Note** that this needs to be tested on a staging server before the final release (so a RC release will be used), since  it touches licenses manager code. I could mainly test `Licensor` code


oanalavinia:
This is not needed for this PR, but I'm adding a comment in case it's needed for future references:
~~WIP:
Functional tests are currently failing. There is some work done, but it's not complete and might need further investigation.~~
-> the functional tests were failing around [here](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/LicensingIT.java#L160), since the UI showed no license available. The actual problem was that the licenses livetable wasn't refreshed (the instance created by docker didn't had the local version of `licensor-ui`, which contained a bug). When I tested manually on the created instance, I falsely assumed that it's not working to add licenses. This _problem_ was reproducing because I needed to execute [UseCertificates](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/LicensingIT.java#L85) (will add the locally created certificates public keys in the static lists from [here](https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicenseValidator.java#L59), which is used for checking validity)

application-licensing-licensor/application-licensing-licensor-api/pom.xml
```diff
@@ -38,6 +38,8 @@
     <xwiki.extension.namespaces>{root}</xwiki.extension.namespaces>
     <checkstyle.suppressions.location>${basedir}/src/main/checkstyle/checkstyle-suppressions.xml</checkstyle.suppressions.location>
     <xwiki.jacoco.instructionRatio>0.50</xwiki.jacoco.instructionRatio>
+    <!-- Added for the licensor-ui dependency, since a JAR extension shouldn't depend on a XAR one. -->
+    <xwiki.enforcer.banneddependencytype-xar.skip>true</xwiki.enforcer.banneddependencytype-xar.skip>
```
trrenty, line 42: Why does it depend on the UI module?

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/MissingLicenseMessageMacro.xml
```diff
@@ -272,7 +284,7 @@
       <contentType>No content</contentType>
     </property>
     <property>
-      <defaultCategory/>
+      <defaultCategories/>
```
trrenty, line 287: I think we should hide this macro, since we only use it for pro apps. Having it in the "Internal" or "Deprecated" category will hide it.

https://www.xwiki.org/xwiki/bin/view/Documentation/AdminGuide/Configuration/#HMacroscategoriesandvisibility

trrenty (commented):
Looks good. Left some comments.

application-licensing-licensor/application-licensing-licensor-api/pom.xml
oanalavinia: From what I understand, this is how the `licensor` was designed, so each pro app has the `api` module as dependency, which has the `ui`. This, as we know, makes the pro (licensed) apps to be identified by the `api` dependency and I guess that they didn't used the `ui` module instead since you might want to depend on it for other usecases. Or simply it wasn't a best practice at that moment to not have apis depending on xar and the current solution felt more natural.
Anyway, I don't think it's an easy fix to avoid it now, maybe only adding also the `ui` dependency besides the `api` one, but I'm not very fond of this. WDYT?

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/MissingLicenseMessageMacro.xml
oanalavinia: I didn't knew about the existing categories. Adding it, thanks!

application-licensing-licensor/application-licensing-licensor-api/pom.xml
trrenty: So, If i understand correctly, the pro apps have a dependency to the `licensor-api` which marks them as an app that requires a license. The `licensor-api` has a dependency to the ui module in order to bring all the xwiki pages when installing the the mentioned pro app.

In order to get rid of this dependency to the ui module, we would need to update the code that checks whether an app depends on the `licensor-api` and update all the pro apps to depend on the `licensor-ui`?

If so, it does sound like plenty of work that might cause some complications. But it might also be fairly simple to do. (unless im missing some implications).

I'm not a fan of also adding the `ui` dependency. Doing that or skipping the enforcer is kind of the same thing - a hack that needs a comment as an explanation.

I think its ok to leave this as is, for now. But it would be useful to have an issue for this.

Issue #174, opened 2024-06-19
New dependencies on licensed extensions are not covered by existing licenses when upgrading licensed extensions


raphj:
When upgrading a licensed extension with a new dependency to another licensed extension, the newly added dependency is not covered by the existing valid license. i.e the existing valid license for the "parent" extension is not considered valid for the dependency.

This caused https://github.com/xwikisas/xwiki-pro-macros/issues/286. The bug can also be reproduced with https://github.com/xwikisas/xwiki-pro-macros/issues/78 (which has gone unnoticed so far).

The dependencies need to be reanalyzed when upgrading extensions, and the new artifact ids need to be added to the valid license.

When handling this, please remove or adjust the workaround added at https://github.com/xwikisas/xwiki-pro-macros/issues/286 (https://github.com/xwikisas/xwiki-pro-macros/commit/c2427420d9b1f78b111116860c8cc0b1a0aa9238) on Pro Macros

mflorea:
The list of dependencies that should be covered by the license are included in the license when it is generated. So if we add paid deps after the license was generated, we currently need to re-generate the license to have the updated list of deps.

On the licensing application side, we could detect when a license needs to be re-generated. E.g.:

* have an extension upgrade listener that catches licensed extension upgrades
* collect the list of licensed deps for both the previous and new extension version (the one being upgraded)
* if the two list differs then make a call to store to regenerate the license then fetch the license from the store and apply it

By re-generate the license I mean: create a new license page with the updated "Dependent Extension Ids" field and with expiration date increased by 1 second. Have a safety measure to block such a request if the previous license was generated very recently (less than X time ago, like 1h or 1 day). The store could also do the same check as the client instance and only re-generate if the deps have changed.

PR #178, opened 2024-10-18
New dependencies on licensed extensions are not covered by existing licenses when upgrading licensed extensions #174 

oanalavinia:


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicenseUpdater.java
```diff
@@ -0,0 +1,249 @@
...
+    private HttpClientUtils httpUtils;
+
+    @Override
+    public void renewLicense(ExtensionId extensionId)
+    {
+        try {
+            logger.debug("Try renewing the license of [{}], in order to include new found changes.", extensionId);
+
+            String errorMsg = String.format(
+                "Failed to update license for [%s]. Please contact sales@xwiki.com for eventual problems.",
```
mflorea, line 124: This looks very similar to the value of the ``ERROR`` static field. Moreover, so far we tried to keep the Licensing Application generic, and put XWiki SAS stuff in the Store Application. Maybe we can keep the message generic or use a translation key whose value can be overwritten from the Store Application.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/LicenseRenewListener.java
```diff
@@ -0,0 +1,208 @@
...
+        licenseUpdater.getLicensesUpdates();
+
+        ExtensionEvent extensionEvent = (ExtensionEvent) event;
+        // Only treat top level licensed extensions. If it's not a top level extension, then its license it's handled
+        // by another extension's license.
+        if (licensedExtensionManager.getLicensedExtensions().contains(extensionEvent.getExtensionId())
+            && isMandatoryLicensedExtension(extensionEvent.getExtensionId()))
+        {
+            InstalledExtension installedExtension = (InstalledExtension) source;
+            if (installedExtension == null) {
```
mflorea, line 110: Would be good to add a comment explaining when the install / upgrade event is triggered without passing the installed / upgraded extension. Looks strange. Also, this check is way faster than the previous one, so I'd check this first.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/LicenseRenewListener.java
```diff
@@ -0,0 +1,208 @@
...
+            licenseUpdater.renewLicense(installedExtension.getId());
+        }
+    }
+
+    private void extensionUpgraded(InstalledExtension installedExtension, Object data, ExtensionEvent extensionEvent,
+        License license)
+    {
+        InstalledExtension prevInstalledExtension =
+            getPreviousInstalledExtension(data, installedExtension, extensionEvent.getNamespace());
+        // Stop if previous version is actually higher, because we don't want to regenerate a license downgrade.
```
mflorea, line 144: Might be good to explain why we don't want to regenerate the license on downgrade (it's not obvious).

Issue #151, opened 2023-09-26
Missing methods in the licensor testing api


trrenty:
When creating integration tests for pro applications, the dependency to the `application-licensing-licensor` module is replaced with a mockup of itself. Currently, only the `hasLicensureForEntity` methods are implemented. If a pro app will use other methods, such as `getLicenseForEntity`, the integration tests will fail with `NoSuchMethodError`.

PR #184, opened 2025-02-12
Missing methods in the licensor testing api #151

trrenty:
Added the missing methods to the script service and to the licensor implementation.


oanalavinia (approved):
Looks good :)

Issue #168, opened 2024-04-08
Include in License section dependencies that are not covered by the license


snazare:
SQ: 
Currently Diagram Pro and PDF Viewer Pro are dependencies of the Pro Macros package as the gliffy, drawio and view-file, view-pdf macros depend on them. If you install Pro Macros and get a license for it, you don't see Diagram Pro and PDF Viewer Pro in the license list, Extension Manager. 

Need: 
Please show in the License section all installed pro apps that require a license, even if they were installed as dependency. 

relates to #171

oanalavinia:
You can split the licensed applications in 2 categories: top level (requires a license) and installed as dependency (it's license will be covered by the license of the extension which installed it).
As this issue states, right now, on the licensing homepage you can only see the top level extensions.

For including all licensed dependencies in the licensing section, some UI changes needed to be made, in order to not bring confusion. The problem would have been for buying / getting a trial license for those extension which were installed as dependencies, because in such cases it is better for users to get the license for the parent extension.

1. A warning was added near the extension name for non-top level extension, with an associated explanation above the table (you can see very quickly why you have that warning). The parent extensions (top-level) are displayed when hovering the name

![image](https://github.com/user-attachments/assets/3fcd619d-654e-4969-b05d-d71d67aa3253)

2. Display a warning when trying to buy / get trial for non-top level extensions

![image](https://github.com/user-attachments/assets/f9d78a60-9f53-42f6-b7d8-cfc55fa24c89)

PR #187, opened 2025-03-21
Include in License section dependencies that are not covered by the license #168

oanalavinia:
* commit partial work
* moved logic to backend, in order to check all transitive dependencies and be able to rule out optional dependencies

To do:
* refactor (maybe move code to another class)
* add unit tests
* check the UI is ok (and refactor it too)


oanalavinia:
For the UI:
You can split the licensed applications in 2 categories: top level (requires a license) and installed as dependency (it's license will be covered by the license of the extension which installed it). 
As this issue states, right now, on the licensing homepage you can only see the top level extensions. 

So, what was done in here is to show all licensed extension, top or installed as dependency. 
In order to differentiate them, a `Installed as dependency` column was added, with some info to mention which extensions depend on it
licensing_homepage

snazare:
Hi @oanalavinia thank you very much for the investigation! I see your point, seeing all the internal modules is not ideal and can cause confusion to the end user. Indeed adding an explanatory message would help and we should also warn users if they try to buy an app installed as dependency. Another idea would be to make the dependencies entries collapsed under the top level application that installed them. This would make the dependencies more clearly separated and related to the top level application. Wdyt? would that be feasible? 

If the collapsed option is complex to implement, we could roll out a fost version with the two warning that you proposed (explanatory message above the livetable + warning when a user tries to buy a license for a dependency)

oanalavinia:
Hi! Thanks for the feedback!
Unfortunately, the collapsed entries were not an option, since livedata does not support it. 

So I made some small changes, hopefully this is more clear.
1. What I did was to remove the extra column (which I had the impression wasn't very visible somehow) and I added a warning near the extension name for non-top level extension, with an associated explanation above the table (you can see very quickly why you have that warning). The parent extensions are displayed when hovering the name
Q: should a more soft icon be used, maybe this warning makes it look like something is wrong?

image

2. Display a warning when trying to buy / get trial for such an extension 

image

PR #194, opened 2025-07-30
Allow oldest users to still use apps which are over the license user limit #204

KebabRonin:
* Added user check so the oldest active users can still use licenses
* User caching
* Tests


application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -135,29 +172,23 @@ public long getUserCount() throws Exception
...
-        statement.append("where doc.fullName = obj.name and obj.className = 'XWiki.XWikiUsers' and ");
-        statement.append("prop.id.id = obj.id and prop.id.name = 'active' and prop.value = '1'");
-
-        Query query = this.queryManager.createQuery(statement.toString(), Query.HQL);
-        query.addFilter(this.countFilter).setWiki(wikiId);
-        List<Long> results = query.execute();
-        return results.get(0);
+        Query query =
+            queryManager.createQuery("object:XWiki.XWikiUsers AND property.XWiki.XWikiUsers.active:true", "solr");
+        query.bindValue("fq", "type:DOCUMENT").bindValue("sort", "creationdate asc");
```
michitux, line 191: It might be interesting to add a `fl` parameter to limit the fields that are retrieved for performance (retrieving just the fields necessary to resolve the reference should be faster than retrieving the full profile information).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -135,29 +172,23 @@ public long getUserCount() throws Exception
...
-        StringBuilder statement = new StringBuilder(", BaseObject as obj, IntegerProperty as prop ");
-        statement.append("where doc.fullName = obj.name and obj.className = 'XWiki.XWikiUsers' and ");
-        statement.append("prop.id.id = obj.id and prop.id.name = 'active' and prop.value = '1'");
-
-        Query query = this.queryManager.createQuery(statement.toString(), Query.HQL);
-        query.addFilter(this.countFilter).setWiki(wikiId);
-        List<Long> results = query.execute();
-        return results.get(0);
+        Query query =
+            queryManager.createQuery("object:XWiki.XWikiUsers AND property.XWiki.XWikiUsers.active:true", "solr");
```
michitux, line 190: It could be interesting to limit the space, e.g., with `space_exact:XWiki` to ensure you don't find user objects in other places.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicenseValidator.java
```diff
@@ -90,6 +95,30 @@ public boolean isSigned(License license)
...
+                try {
+                    allowedUsers = userCounter.getOldestUsers((int) license.getMaxUserCount());
+                } catch (Exception e) {
+                    this.logger.warn(
+                        "Failed to check the allowed users. Assuming the license is not valid. Root cause is: [{}].",
+                        ExceptionUtils.getRootCauseMessage(e));
+                    allowedUsers = List.of();
+                }
+                this.logger.debug("Allowed users for license [{}]: [{}].", license, allowedUsers);
+                return allowedUsers.contains(context.get().getUserReference());
```
michitux, line 115: For performance, it would be much better to use (and cache) a `Set` here to avoid going through the full list of all users. In particular, as I expect that this code could be executed frequently on a single page. Also, what about guests/public access?

KebabRonin: Another option could be to store 'marker' users at the license maxUser limits, and then compare the creationDate of the current user with the 'marker' user's creationDate. Could also store the neighboring +-5 users to the 'marker' to not invalidate the cache too much when adding/removing a couple of users.

So if the 100-th user is `u100`, only store him instead of 100 users, and compare the creation date to determine who has access to a license.
If a wiki has multiple licenses, the licensor would only store the users at the maxUser thresholds, like `u10`, `u100`, `u100000`.

I was also thinking of updating the cache manually when a new `XWiki.XWikiUsers` object is added, like adding +1 to the count and updating the 'marker' users. But I don't know if it's that useful, especially since it will probably introduce some desync bugs.  

For guest users/logged out access, the missing license message will appear. It would be easy to change, but I don't know what the intended behavior should be. I assumed that it would be weird if you could view/use an app if you are logged out (like ideas app with anonymous voting), but have a missing license if you have an account.

michitux: I'm not sure regarding this marker user strategy, you would update the marker user when users are deactivated/activated or deleted? In any case, if you don't have that yet (I didn't check), you'll need some cache invalidation when users are removed or deleted. You should take into account there that Solr indexing can be delayed, I'm not sure actually how that could work. Maybe relying on the database for such a crucial feature might be better in the end.

Regarding guests, did you test this with authenticator extensions? Is it still possible with them to login (where you're first guest) after the license limit has been exceeded? To me, it makes sense to allow guests also when the user limit is exceeded, but I'm not the one who decides this (in particular for public websites).

Just as an idea, an alternative to allowing some users would be to have a time-based grace period. Like to say we display warnings for everybody and give a grace period of, e.g., 14 days to either disable/delete users or extend the license.

KebabRonin:
Adding the exception for guest users to have a license when user limit is exceeded fixes the Active Directory issue of users not being able to log in at all.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/enforcer/UserLicenseSecurityCacheInvalidator.java
```diff
@@ -0,0 +1,114 @@
...
+    {
+        super(ROLE_NAME,
+            Arrays.asList(new DocumentCreatedEvent(XWIKI_SPACE_FILTER), new DocumentUpdatedEvent(XWIKI_SPACE_FILTER),
+                new DocumentDeletedEvent(XWIKI_SPACE_FILTER)));
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        // Only invalidate security cache when users are modified.
```
KebabRonin, line 97: Security cache breaks when the license status changes for a user, because another user was (de)activated and the user limit was exceeded.

Tested with Task Manager Pro app:
1. For a license of 2 users for the Task Manager Pro app
2. Create u1 and u2 users, u1 created first
3. Admin and u1 should have access to the app, u2 not
4. Disable u1
5. Without security cache invalidation, u2 will see an empty `TaskManager.WebHome`, because the license check returns `true`, but the security cache still denies u2's view rights for `WebHomeSheet`

PR #207, opened 2025-09-30
Do not block access for ALL users when the license user limit is exceeded #192

### LINKED ISSUE #192: Do not block access for ALL users when the license user limit is exceeded
**Is your feature request related to a problem? Please describe or add steps to reproduce your use case.**

- Currently, if, in the wiki the number of existing users is higher than the license user limit and there's an AD Pro active license, then _all_ the users are all blocked out of the wiki and can't log in (so when there's at least one user created over the license user limit)

**Describe the solution you'd like**

- The license should always function for the users already created before the license user limit was exceeded.  
- Not all users should be blocked, the existing users should still be able to login. Only users that are were created after the license users limit was reached (until a potential user limit increase or until other users are disabled from the wiki so that the license user limit is no _longer_ exceeded)

KebabRonin:
Mostly API changes to better support this functionality in Auth Extensions.

- _`DefaultLicensor.java`_, _`DefaultLicenseManager.java`_ - Some additional APIs which accept extension IDs as `String` instead of `ExtensionId`
- _`UserCounter.java`_ - Added an in-memory cache of active (non-disabled) users on the instance, storing the users in the order of their creation date
- _`AuthExtensionUserManager.java`_ - Added interface for auth extensions, with mostly unused methods


KebabRonin:
### How to store the users managed by an Auth extension?
Each Auth extension will probably need to implement a component which describes which users are managed by an app, and which users are still allowed after a license expiration.
#### Solr
PRO:
+ Can take advantage of custom indexes
+ Should be faster and more efficient

CON:
- Depends on unpredictable (re)indexing operations - could have unexpected results until the indexing is done and users' active statuses are recomputed
- Has a bit of lag after a user change before the solr query reflects the real state of the DB, since indexes must be recomputed
#### DB (HQL)
PRO:
+ Information always(?) reflects reality 

CON:
- Need to run the same query on all subwikis
- Could be slow
#### In-Memory cache
PRO:
+ Easy access from java

CON:
- Storing all users in memory may become costly for large instances - even if the cache is stored on disk when not in use
- Need to implement cache updates/invalidation

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -54,6 +62,11 @@
 @Singleton
 public class UserCounter
 {
+    protected static final String BASE_USER_QUERY = ", BaseObject as obj, IntegerProperty as prop "
```
KebabRonin: Protected only for being accessible from the `UserCounterTest` class.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -73,6 +86,12 @@ public class UserCounter
 
     private Long cachedUserCount;
 
+    // A set of users on the instance, sorted by creation date.
+    private SortedSet<XWikiDocument> cachedSortedUsers;
+
+    // Helper to find users in constant time.
```
KebabRonin: Since the `cachedSortedUsers` is populated from the database, dummy `XWikiDocument`s can't be used to index into the set directly (equals fails because it checks for a bunch of fields like version, author, etc.).

This is the reason why the map is needed.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -124,11 +145,41 @@ public void onEvent(Event event, Object source, Object data)
...
+     *
+     * @return the users, sorted by creation date.
+     */
+    public SortedSet<XWikiDocument> getSortedUsers() throws WikiManagerException, QueryException
+    {
+        if (cachedSortedUsers == null) {
+            cachedSortedUsers = new TreeSet<>(
+                (e1, e2) -> new CompareToBuilder().append(e1.getCreationDate(), e2.getCreationDate()).build());
+            for (String wikiId : wikiDescriptorManager.getAllIds()) {
+                cachedSortedUsers.addAll(getUsersOnWiki(wikiId));
```
oanalavinia: So we retain in this variable all the users from all the wikis. The use case is that you have a license for a number of users, this being applied per instance. If your license is not valid anymore because of too many users, we will keep the first x created users, no matter the wiki they are part of. Right?

oanalavinia (commented):
I added some comments. Some are just for small things, but there are also questions related to the functionality / architecture. I will continue the review once those are clarified, since some comments might not apply anymore after (I think it's the case now as well)

I had another question related to this functionality. Will this affect other authenticators? For example, if I also have EntraID app installed, what's the behavior?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/AuthExtensionUserManager.java
```diff
@@ -0,0 +1,81 @@
...
+import org.xwiki.model.reference.DocumentReference;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+/**
+ * Interface for extensions which manage users, so the license user limit can only take into account the users managed
+ * by those extensions for computing the user limit.
+ * <p>
+ * Components implementing this interface should be named as extension id, so the user manager is discoverable.
```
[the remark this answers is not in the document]
oanalavinia: > I was thinking to have an AuthExtensionUserManager named active-directory-api for AD, azure-oidc for OIDC, etc, so the users can be associated with an extension and be able to use the extension id to get the associated user manager.

I was saying to add a small example (as you mentioned `MyExtensionAuthUserManager` here) in the code

But, given the discussion from https://github.com/xwikisas/application-licensing/pull/207#discussion_r2726963007 , this class is nt actually needed anymore, right? Or would you like to use it for the user warning?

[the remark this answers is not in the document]
KebabRonin: Just altering the login message from the extension should be enough to make it less confusing to users who can't log in anymore.

I think the class can be removed for now, and if the need arises to do something similar (like listing affected users when an extension is disabled) in the future, this can be a starting point.

oanalavinia (changes requested):

application-licensing-test/application-licensing-test-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Component used to count the existing active users.
+ * 
+ * @version $Id$
+ * @since 1.6
+ */
+@Component(roles = UserCounter.class)
+@Singleton
+public class UserCounter
```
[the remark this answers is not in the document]
oanalavinia: Ok, you could add a comment

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -46,14 +52,19 @@
...
+ *
  * @version $Id$
  * @since 1.6
  */
 @Component(roles = UserCounter.class)
 @Singleton
 public class UserCounter
 {
+    protected static final String BASE_USER_QUERY =
+        ", BaseObject as obj, IntegerProperty as prop " + "where doc.space = 'XWiki' "
```
oanalavinia: > "where doc.space = 'XWiki' "

Why do you need it?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -73,6 +84,12 @@ public class UserCounter
 
     private Long cachedUserCount;
 
+    // A set of users on the instance, sorted by creation date.
+    private SortedSet<XWikiDocument> cachedSortedUsers;
+
+    // Helper to find users in constant time.
+    private Map<DocumentReference, XWikiDocument> cachedSortedUsersLookupTable;
```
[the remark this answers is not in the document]
abrassat: Yes, I'll add it on the next commit.

I was hesitant because it looks like a time/memory tradeoff, but I don't know for sure the impact on one or the other without a benchmark.

oanalavinia (approved):
This looks ok on my side, let me know if maybe you are not sure of some aspects. Thanks for the work!

Small note that the performance tests are still to be done + there is still https://github.com/xwikisas/application-activedirectory/pull/112 which is related

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/AllIT.java
```diff
@@ -70,4 +70,11 @@ class AllIT
     class NestedLicensingIT extends LicensingIT
     {
     }
+
```
oanalavinia, line 79: I just realized. Could you add functional tests for the cases you added as part for this PR? You would test all the cases you mentioned, if I can login with a older user, but with one from when the limit was reached, etc. . WDYT?

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/AllIT.java
```diff
@@ -70,4 +70,11 @@ class AllIT
...
+    // This test takes ~40 min to set up. It should only run when UserCounter class is updated.
```
oanalavinia: I guess it's ok to keep it commented out and have a procedure for when it should be executed. Can you add a little bit more context to this comment? Also, please format the added code

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserStressTestLicensingIT.java
```diff
@@ -0,0 +1,90 @@
...
+class UserStressTestLicensingIT
+{
+    private static final String PASSWORD = "password";
+
+    private static final int USERS_COUNT = 10_000;
+
+    // The tested user limit MUST be lower than the amount of users on the instance, otherwise the method returns
+    // TRUE without actually populating the user list.
+    private static final String TEST_PAGE_SCRIPT =
+        "{{velocity}}\n" + "$services.licensing.licensor.isCurrentUserUnderLimit($xwiki.getUser().getUser()"
```
oanalavinia: Isn't the method name `isUserUnderLimit`?

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserStressTestLicensingIT.java
```diff
@@ -0,0 +1,90 @@
...
+            + ".getUserReference(), " + (USERS_COUNT - 100) + ")\n" + "{{/velocity}}";
```
oanalavinia: Can you arrange it a little bit? E.g. https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java#L503

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserStressTestLicensingIT.java
```diff
@@ -0,0 +1,90 @@
...
+ * Functional tests for the Licensing application.
+ *
+ * @version $Id$
+ */
+@UITest
+class UserStressTestLicensingIT
+{
+    private static final String PASSWORD = "password";
+
+    private static final int USERS_COUNT = 10_000;
```
oanalavinia:
```suggestion
    private static final int USER_COUNT = 10_000;
```

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserStressTestLicensingIT.java
```diff
@@ -0,0 +1,90 @@
...
+    void stressTest(TestReference testReference, TestUtils setup)
+    {
+        setup.loginAsSuperAdmin();
+        // A lot of users on the test instance shouldn't be crashing the licensor when it tries to compute the user
+        // limit.
+        setup.createPage(testPage, TEST_PAGE_SCRIPT);
+        setup.gotoPage(testPage, "view");
+        setup.gotoPage(testPage, "view");
+        assertTrue(setup.isInViewMode());
+        // We didn't crash (probably).
```
oanalavinia: The `probably` part doesn't seem ok to be added as comment.
If you are not sure if this check is enough, I would actually do more steps, e.g. login with some user, check again the script page, check one without script, then logout

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserStressTestLicensingIT.java
```diff
@@ -0,0 +1,90 @@
...
+import org.xwiki.test.docker.junit5.UITest;
+import org.xwiki.test.ui.TestUtils;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Functional tests for the Licensing application.
+ *
+ * @version $Id$
+ */
```
oanalavinia, line 41: Needs an update

oanalavinia (changes requested):

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/AllIT.java
abrassat: I'm not sure all those test cases would apply here, as the licensor doesn't actually block the login.

I can add tests to see if the user is correctly identified to be under the limit, but the rest of the logic is handled in AD

(AD is lacking tests entirely, and would need to do some digging to find a mock config for running AD in the docker tests)

oanalavinia: I though that's the case (that AD is actually the one using that underTheLimit logic)

Maybe we can mock it in licensing for testing purposes? Meaning that we have to find a way to test it even if it's not a real flow.
For example, you have a script page where you do something similar as in the StressTest page, where you check if the current user is under the limit. Then you create the user structure where some users are still valid and some are not. The test would be to access the script page with each user and check if `isUserUnderLimit` returns the right value. Would it be possible?

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/script/LicensorScriptService.java
```diff
@@ -299,6 +299,26 @@ public Long getUserCount()
         }
     }
 
+    /**
+     * List the users on the entire instance, ordered by creation date. If the passed user is at an index below the
+     * limit, return true.
+     * @param user the user to test
```
oanalavinia, line 306: formating missing

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,308 @@
...
+    {
+        // Make sure we're on the right subwiki.
+        String currentWiki = setup.getCurrentWiki();
+        setup.setCurrentWiki(testPage.getWikiReference().getName());
+
+        setup.gotoPage(testPage, "view", Map.of("user", user, "limit", limit));
+        String pageSource = setup.getDriver().findElement(By.id("xwikicontent")).getText();
+
+        setup.setCurrentWiki(currentWiki);
+        return pageSource.startsWith("true");
```
oanalavinia: You could use a html structure (or wiki syntax) and retrieve this value with `getElementById`

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/script/LicensorScriptService.java
```diff
@@ -299,6 +299,26 @@ public Long getUserCount()
...
+     * @return true if the passed user is below the specified limit, false otherwise
+     * @since 1.33.0
+     */
+    @Unstable
+    public boolean isUserUnderLimit(DocumentReference user, long limit)
+    {
+        try {
+            return this.userCounter.isUserUnderLimit(user, limit);
+        } catch (Exception e) {
+            this.logger.warn("Failed to determine if the user is under the license user limit. Root cause is: [{}].",
```
oanalavinia:
```suggestion
            this.logger.warn("Failed to determine if user is under the license user limit. Root cause is: [{}].",
```

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -207,9 +212,8 @@ public boolean isUserUnderLimit(DocumentReference user, long userLimit) throws E
         if (null == user) {
             return false;
         }
-        if (userLimit < 0 || getUserCount() <= userLimit) {
-            // Unlimited licenses should always return true.
-            // Also, skip the checks for instances with fewer users than the limit.
```
oanalavinia: Why don't we do this anymore?

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,308 @@
...
+    {
+        setup.loginAsSuperAdmin();
+        setup.createPage(testPage, TEST_PAGE_SCRIPT);
+    }
+
+    int getUserCountOnInstance(TestUtils setup) throws Exception
+    {
+        setup.gotoPage(testPage, "view");
+        String pageSource = setup.getDriver().findElement(By.id("xwikicontent")).getText();
+        return Integer.parseInt(pageSource.split("\n")[1]);
```
oanalavinia: Same as for the other, I would use a html structure in order to be able to use `getElementById`

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,308 @@
...
+
+        deleteUsers(setup, "xwiki", 1, 3);
+        deleteUsers(setup, subwikiName, 3, 6);
+        deleteUsers(setup, "xwiki", 6, 7);
+    }
+
+    /**
+     * Get the result of a isUserUnderLimit() call, by evaluating it in a velocity script.
+     */
+    private boolean checkIsUserUnderLimit(TestUtils setup, String user, int limit)
```
oanalavinia: Semantically, this is more accurate as you respond with true / false

```suggestion
    private boolean isUserUnderLimit(TestUtils setup, String user, int limit)
```

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,308 @@
...
+        Assertions.assertTrue(checkIsUserUnderLimit(setup, "xwiki:XWiki.User_2", 5));
+
+        Assertions.assertEquals(2, this.getUserCountOnInstance(setup));
+
+        deleteUsers(setup, "xwiki", 1, 3);
+    }
+
+    /**
+     * Test that the isUserUnderLimit function returns true for existing users, as long as the number of users on the
+     * instance is lower than the requested limit.
```
oanalavinia: Can you update the comments (here and the others) to specify the current use case. The current comment could be used for the whole class. e.g. Verify that users over the limit are marked correctly

oanalavinia (changes requested):

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
vmassol: Note that this is not following the PageObject strategy (the quick to know if you're following it or not, is that there should not be any call to getDriver() inside a Test class. That needs to go in the PO class).

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
abrassat: If the test is skipped when the user count is less than the limit, then you can get the result that inexistent or disabled users are considered to be under the limit (even if they logically shouldn't).

Another option would be to move this test below, and only execute `sortedSet.headSet().size()` if the `sortedSet.size()` is under the limit.

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/UserCounter.java
```diff
@@ -223,7 +223,12 @@ public boolean isUserUnderLimit(DocumentReference user, long userLimit) throws E
         if (userDocument == null) {
             return false;
         } else {
-            return sortedUsers.headSet(userDocument).size() < userLimit;
+            // Only build the headSet if the number of users on the instance is over the limit.
+            if (sortedUsers.size() <= userLimit) {
```
abrassat, line 227: This `if` is why I made `moreThanRequested` and `lowerThanRequested` tests. I'm not sure if it's really needed though.

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
oanalavinia: You need to do more than simply use `new ViewPage().getContent()`. It might feel like you solved the PageObject strategy comment, but there is still the one about using a proper structure, rather than relying on hacks like `return Integer.parseInt(pageSource.split("\n")[1])`

To give you an idea, you would have a UserCounterPage, on which you can use `UserCounterPage#getCount`
On that page, you would have `<span id="userCount">$services.licensing.licensor.getUserCount()</span>` , so `#getCount` is actually using `getElementById`

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * @version $Id$
+ * @since 1.33.0
+ */
+@UITest
+class UserCounterIT
+{
+    private static final String PASSWORD = "password";
+
+    private static final String TEST_PAGE_SCRIPT =
```
oanalavinia: Since this is also starting to have more content and an actual structure, you can use a vm template (e.g. https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-security/xwiki-platform-extension-security-test/xwiki-platform-extension-security-test-docker/src/test/it/org/xwiki/extension/security/test/ui/ExtensionSecurityIT.java#L89 )

abrassat: I used the same method of including the vm template as https://github.com/xwikisas/application-licensing/blob/8678a7264decf39712b2041f3c54efc4d96f5899/application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/LicensingIT.java#L75

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+    public static UserCounterTestPage gotoPage(TestUtils setup, String user, Integer limit) throws Exception
+    {
+        // Make sure we're on the right subwiki.
+        String currentWiki = setup.getCurrentWiki();
+        setup.setCurrentWiki(UserCounterTestPage.testPage.getWikiReference().getName());
+
+        // Create the page if it doesn't exist.
+        try (InputStream content = UserCounterTestPage.class.getResourceAsStream("/userCounterTestPage.wiki")) {
+            setup.rest().savePage(testPage, IOUtils.toString(content, StandardCharsets.UTF_8), "");
+        }
```
oanalavinia: Why are you doing this in here, instead of creating this page in the test setup method?

oanalavinia: So you have this page on the main wiki only, this is why you need to play with the current wiki

Could you maybe have this page created on the current wiki (being it main wiki or sub wiki), and simply navigate to that?
I would also move the actual creation outside the page object, in the test

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+
+import org.apache.commons.io.IOUtils;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.test.ui.TestUtils;
+import org.xwiki.test.ui.po.ViewPage;
+
+public class UserCounterTestPage extends ViewPage
```
oanalavinia:
```suggestion
public class UserCounterPage extends ViewPage
```

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+    private WebElement isUserUnderLimit;
+
+    public static UserCounterTestPage gotoPage(TestUtils setup, String user, Integer limit) throws Exception
+    {
+        // Make sure we're on the right subwiki.
+        String currentWiki = setup.getCurrentWiki();
+        setup.setCurrentWiki(UserCounterTestPage.testPage.getWikiReference().getName());
+
+        // Create the page if it doesn't exist.
+        try (InputStream content = UserCounterTestPage.class.getResourceAsStream("/userCounterTestPage.wiki")) {
```
oanalavinia:
```suggestion
        try (InputStream content = UserCounterTestPage.class.getResourceAsStream("/UserCounterTestPage.wiki")) {
```

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+    {
+        setup.deletePage(testPage);
+    }
+
+    public int getUserCount()
+    {
+        return Integer.parseInt(userCount.getText());
+    }
+
+    public boolean getIsUserUnderLimit()
```
oanalavinia:
```suggestion
    public boolean isUserUnderLimit()
```

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public int getLimitParam()
+    {
+        return Integer.parseInt(limitParam.getText());
+    }
+
+    public String getUserParam()
+    {
+        return userParam.getText();
+    }
```
oanalavinia: Are these used?

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        if (limit != null) {
+            queryParams.put("limit", limit);
+        }
+        setup.gotoPage(testPage, "view", queryParams);
+        setup.setCurrentWiki(currentWiki);
+        return new UserCounterTestPage();
+    }
+
+    public static void deletePage(TestUtils setup)
```
oanalavinia: I would move this in the test

application-licensing-test/application-licensing-test-docker/src/test/resources/userCounterTestPage.wiki
```diff
@@ -0,0 +1,15 @@
+{{velocity}}
+{{html}}
+isUserUnderLimit:
```
oanalavinia:
```suggestion
Is user under limit:
```
Same for the others

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,284 @@
...
+        Assertions.assertFalse(isUserUnderLimit(setup, "xwiki:XWiki.User_6", 4));
+
+        Assertions.assertEquals(6, this.getUserCountOnInstance(setup));
+
+        deleteUsers(setup, "xwiki", 1, 3);
+        deleteUsers(setup, subwikiName, 3, 6);
+        deleteUsers(setup, "xwiki", 6, 7);
+    }
+
+    int getUserCountOnInstance(TestUtils setup) throws Exception
```
oanalavinia:
```suggestion
    private int getUserCountOnInstance(TestUtils setup) throws Exception
```

oanalavinia (changes requested):

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterTestPage.java
abrassat: These functions are not used. I added information about the query parameters for easier debugging when viewing the failed test recording.

I'll remove these functions, and leave that information only in the page, since the query parameters should be known when the page is visited in the test.

abrassat:
I noticed that, on my setup, the `LicensingIT.generateLicense()` test fails due to missing superadmin Programming rights on `LicensesPublic.Code.LicensesNotificationsUIX`. (This happens on both this branch and `xwikisas:master` too)

That page should probably(?) be included in this regex: (as a misc commit on master)
https://github.com/xwikisas/application-licensing/blob/8678a7264decf39712b2041f3c54efc4d96f5899/application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/AllIT.java#L38

application-licensing-test/application-licensing-test-pageobjects/src/main/java/com/xwiki/licensing/test/po/UserCounterPage.java
```diff
@@ -0,0 +1,43 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.licensing.test.po;
+
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+import org.xwiki.test.ui.po.ViewPage;
+
+public class UserCounterPage extends ViewPage
```
oanalavinia, line 32: Comment missing for this class

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,315 @@
...
+    @Test
+    void otherUserTest(TestUtils setup) throws Exception
+    {
+        // 0 users on the instance for these tests.
+        Assertions.assertEquals(0, this.getUserCountOnInstance(setup));
+
+        Assertions.assertFalse(isUserUnderLimit(setup, "xwiki:XWiki.XWikiGuest", 1));
+        Assertions.assertFalse(isUserUnderLimit(setup, "xwiki:XWiki.superadmin", 1));
+        Assertions.assertFalse(isUserUnderLimit(setup, "xwiki:XWiki.NonExistentUser", 1));
+        // Unlimited limit.
```
oanalavinia:
```suggestion
        // A -1 limit means the license has no limit.
```

application-licensing-test/application-licensing-test-docker/src/test/it/com/xwiki/licensing/test/ui/UserCounterIT.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    /**
+     * This test documents the behavior of isUserUnderLimit() for special users.
+     */
+    @Test
+    void otherUserTest(TestUtils setup) throws Exception
+    {
+        // 0 users on the instance for these tests.
+        Assertions.assertEquals(0, this.getUserCountOnInstance(setup));
```
oanalavinia: I would import these, to not need to mention `Assertions` every time

```
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertFalse;
import static org.junit.jupiter.api.Assertions.assertTrue;
```

oanalavinia (approved):
I would say it looks ok overall. Just some small remarks

The next steps would be:
1. Release an RC version
2. Make a AD release to include this licensor version and https://github.com/xwikisas/application-activedirectory/pull/112
3. Have it tested to make sure there are no regressions

oanalavinia:
@abrassat forgot to mention that this issue should be merged on a feature branch, in order to not block the main branch while we test and confirm everything works as expected

Issue #212, opened 2025-10-26
There is a typo inside the DefaultLicensedExtensionManager


Farcasut:
In the log there is 'conseuence' instead of 'consequence'
https://github.com/xwikisas/application-licensing/blob/a8fee8ef14afd2b47bf5b31f640ca85d8ba0615e/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/DefaultLicensedExtensionManager.java#L120

PR #215, opened 2025-11-10
Fix typos from application licensor

KebabRonin:
Would it be ok to include other typos/grammar issues in java comments in this PR?

Edit: Originally opened for #212


ChiuchiuSorin:
I think it would be ok to include more typos, but I'd rename the issue to something like `Fix typos from application licensor`.

Issue #208, opened 2025-10-07
Add configuration for who should get licensing notifications


oanalavinia:
**Is your feature request related to a problem? Please describe or add steps to reproduce your use case.**
This issue is related to the licensing notifications you get as an administrator. For the type of notifications you can get check https://github.com/xwikisas/application-licensing/issues/127 and https://github.com/xwikisas/application-licensing/issues/110 (not sure if we have others).

The problem is that these notifications are send to administrators, while on some instances these users are not actually used, at least not that much. On this cases the notification can be lost

**Describe the solution you'd like**
I would like to propose to have a configuration in which you can set a specific group which should receive these notifications. Like this, you can have a configurable group of users which will receive the notifications without making them admins

ChiuchiuSorin:
Added a new configuration option at the bottom of the licenses section in administration, where an admin will be able to select which groups will receive notifications. The admin group is added by default:
<img width="1800" height="900" alt="Image" src="https://github.com/user-attachments/assets/e97230e3-a07e-4a9c-af01-d46f6fa2932e" />

Modified how licenses notifications are displayed for non admin users:
<img width="692" height="334" alt="Image" src="https://github.com/user-attachments/assets/c2af87ec-348a-4960-a53b-3ab69ce1faef" />

PR #216, opened 2025-11-10
Add configuration for who should get licensing notifications #208

ChiuchiuSorin:
Added a new `LicensingNotifications` class that manages the target groups who need to receive notifications like expired/expiring licenses, new application version or automatic application upgrade events notifications.

UI changes:
  * added a group selector in the `Licenses` section:
 Screenshot from 2025-11-10 13-49-06


application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/LicensesNotificationsUIX.xml
```diff
@@ -483,17 +483,27 @@
...
-      <async_enabled/>
+      <async_enabled>0</async_enabled>
     </property>
     <property>
       <content>{{velocity}}
-#if ($services.security.authorization.hasAccess('view', 'Licenses.Code.LicensesNotificationsUIX'))
+#set ($licensingConfigDoc = $xwiki.getDocument('Licenses.Code.LicensingConfig'))
+#set ($notificationClass = 'Licenses.Code.LicensingNotificationClass')
+#set ($notificationsObj = $licensingConfigDoc.getObject($notificationClass))
+#set ($notifiedGroups = $notificationsObj.getValue('notifiedGroups').split(','))
```
oanalavinia: You could expose `getNotifiedGroupsSet` through the licensing script service

application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/script/LicensorScriptService.java
```diff
@@ -107,6 +111,20 @@ public void initialize() throws InitializationException
...
+     *
+     * @return {@link List} with the groups whose members need to be notified about a licensor event
+     * @throws AccessDeniedException if the requesting user lacks admin rights
+     * @since 1.31
+     */
+    @Unstable
+    public List<String> getNotificationTargetGroups() throws AccessDeniedException
+    {
+        contextualAuthorizationManager.checkAccess(Right.ADMIN);
+        return licensingConfig.getNotifiedGroups();
```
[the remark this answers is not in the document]
ChiuchiuSorin: > So if I add here a group which contain users without this rights, they won't actually see the notification?
I think this property is redundant if that check is not updated too. For example, every user with admin rights should be in XWikiAdminGroup anyway, so adding here another group is not doing anything. Of course, there are those instances where they don't use the normal groups.

Considering your mentions here, if we are to only rely on group checks, I'd also have to remove the admin access check from this function. To avoid exposing the API to unauthorized users, I'll only return a boolean with the value true if the user is part of the notified groups, or false otherwise.

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/VelocityMacros.xml
```diff
@@ -46,25 +46,7 @@
 #set ($userCount = $services.licensing.licensor.userCount)
 #set ($notificationClass = 'Licenses.Code.LicensingNotificationClass')
 #includeMacros("XWiki.AdminFieldsDisplaySheet")
-
-#macro (getPaidExtensions)
```
oanalavinia: Why did you moved this macros in `LicensesNotificationsUIX`? It makes more sense for them to be here

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/WebHome.xml
```diff
@@ -108,6 +108,7 @@
 #else
   #set ($discard = $xwiki.ssx.use('Licenses.WebHome'))
   #set ($discard = $xwiki.jsx.use('Licenses.WebHome'))
+  #displayNotificationsConfiguration
```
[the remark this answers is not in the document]
oanalavinia: Adding that break would not help also if you add this configuration at the end?

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/VelocityMacros.xml
oanalavinia: This means that, as a non-admin which is in the notification group, I will get a notification that, for example, licenses have expired / are going to expire, but I won't be able to access the licenses page which is in the notification, right? If this is the case, the notification is misleading and it would look like a bug

ChiuchiuSorin:
Notification message for non admin users:
image

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/Translations.xml
```diff
@@ -90,13 +90,19 @@ licensor.ownerDetails.instanceId.hint=Unique identifier for your XWiki instance.
...
 licensor.licenseManager.installedAsDependency.hint=Extensions marked by this sign were installed as dependencies of other extensions and their license will be covered by the parent license
 licensor.expiringLicenses=The following extensions are about to expire:
 licensor.expiringLicenses.warning=Check licenses
+licensor.expiringLicenses.notifyAdmin=To avoid breaking functionality, please notify an admin to extend the license.
 licensor.missingLicenses=The following extensions do not have a valid license:
 licensor.missingLicenses.danger=Renew licenses
+licensor.missingLicenses.notifyAdmin=For this to be fixed, please notify an admin.
```
oanalavinia:
```suggestion
licensor.missingLicenses.notifyAdmin=Please contact an administrator for extending the license.
```

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/Translations.xml
```diff
@@ -90,13 +90,19 @@ licensor.ownerDetails.instanceId.hint=Unique identifier for your XWiki instance.
...
+licensor.notificationSettings.heading=Notification settings
+licensor.notificationSettings.hint=Manage licenses notifications
 licensor.licenseManager.heading=Licensed Extensions
 licensor.licenseManager.hint=Here's a list of extensions installed in your wiki that require a license to be activated. You can buy a license or get a 10-day trial license that can be extended only once with 10 days. A license covers all versions of the licensed extension and is shared by the main wiki and all its subwikis. A licensed extension is listed multiple times if different versions are installed on different subwikis, but all those versions will share the same license. For each extension you can also enable the automatic upgrade to the lastest available stable version.
 licensor.licenseManager.installedAsDependency.hint=Extensions marked by this sign were installed as dependencies of other extensions and their license will be covered by the parent license
 licensor.expiringLicenses=The following extensions are about to expire:
 licensor.expiringLicenses.warning=Check licenses
+licensor.expiringLicenses.notifyAdmin=To avoid breaking functionality, please notify an admin to extend the license.
```
oanalavinia:
```suggestion
licensor.expiringLicenses.notifyAdmin=Please contact an administrator for extending the license.
```

application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/VelocityMacros.xml
oanalavinia: > I moved them as the rights to view/edit on the Licenses space are restricted to the Admin group. I gave the rights to view for the XWikiAllGroup on LicensesNotificationsUIX (to display the notifications for any user, no matter the rights, as long as is a memeber of a group set in administration), but I cannot use the macros from the Velocitymacros, as those are restricted by the Licenses space restrictions.

Back on the initial problem, I still don't like that you have to move generic macros in a page for licenses notification, since it's not nice to have to include that `#includeMacros("Licenses.Code.LicensesNotificationsUIX")` . Now you have also moved `getLicensesAdminSectionURL`, so this page has a lot of macros. Maybe you could also open the VelocityMacros page, or have another page with just some macros, and only have those opened? Not sure what else there is on the VelocityMacros page, maybe second option is better

oanalavinia (approved):
Looks ok, thanks!

Issue #40, opened 2018-04-24
Choosing the best license based on expiration date limits the ability to renew a license


mflorea:
I have this use case:

* the user gets a license for Silver support with 100 user limit that expires in 2019
* the user then extends the license with one year. The new license expires in 2020
* the user then wants to change the plan to Gold support with 250 user limit. The remaining days are converted to the new plan and the new license expires sooner than the existing license.

We end up having two licenses that are hard to compare:

* the first license has a better expiration date but the constraints (support and user limit) are worse
* the second license has better constraints but worse expiration date

ATM from the licensing application point of view the first license wins. But this is not the expected behavior. For this use case the most recent license should win, even if it has a less good expiration date, but the license API doesn't store the date when the license was generated.

PR #222, opened 2026-01-22
Choosing the best license based on expiration date limits the ability to renew a license #40

rbancescu:


application-licensing-common/application-licensing-common-api/src/main/java/com/xwiki/licensing/License.java
```diff
@@ -326,6 +331,24 @@ public static License getOptimumLicense(License license1, License license2)
...
+
+        if (!license1HasIssueDate && license2HasIssueDate) {
+            return license2;
+        }
+
+        if (license1HasIssueDate && license2HasIssueDate) {
+            long license1IssueDate = Long.parseLong(license1.getLicensee().get(LICENSE_ISSUE_DATE));
+            long license2IssueDate = Long.parseLong(license2.getLicensee().get(LICENSE_ISSUE_DATE));
+
+            return license1IssueDate > license2IssueDate ? license1 : license2;
```
oanalavinia: You are now choosing the license with the best issue date, but I would still check for it to not be expired. Or is it done before?

raresbancescu: The check is done before comparing the licenses; In this method only the non expired licenses are checked!

oanalavinia (changes requested):
I was also checking this schema https://github.com/xwikisas/application-licensing/blob/master/application-licensing-common/application-licensing-common-model/src/main/resources/xwiki.license.model.xsd
Could you check if there isn't a need to update it with the new field?

PR #233, opened 2026-05-06
Newly requested trial licenses invalidate the existing license #228

### LINKED ISSUE #228: Newly requested trial licenses invalidate the existing license
**Describe the bug**
When requesting a trial license, a new license is generated at that exact time. With the change brought in by 
https://github.com/xwikisas/application-licensing/issues/40#issuecomment-3906948740, it makes this new license get applied and it overwrites the old one.

The change mentioned above is welcomed, but it seems it needs more polish. By mistake, any admin could request a trial license and invalidate the old existing license. For example:
* XWiki user has XWiki Pro
* XWiki user wants to test Confluence Migration Tookit and they request a trial license
* The new generated license overwrites the old XWiki Pro license and all the paid extensions are controlled by the trial license.

rbancescu:


oanalavinia (approved):

oanalavinia:
We checked also why the license was extended starting from the trial license expiration date, instead of the current date. 
The logic is explained in this (old) [issue](https://jira.xwikisas.com/browse/LICENSING-45) : the button was always displayed and after the 2 usages it wasn't producing any effects.

Since we aren't in that case anymore, computing now the second expiration date from the current date shouldn't have any side effects

@rbancescu would be good to mention this on the parent ticket when you merge the fix

